Issue #227, opened 2013-11-15
Does angular translate support right-to-left?


jameshines10:
Does angular translate support right-to-left? If not, what is the best way to support right-to-left languages in angular translate?

zeeshanjan82:
any updates whether angular-translate supports right to left languages?

tspaeth:
@zeeshanjan82 Please look into #260 - and we're happy to get any PR or concept work on this!

PR #262, opened 2013-12-18
Add predefined namespace for storing language-specific information

DWand:
Here is a little proposal which, probably, could be useful to solve #227 and to provide new features in the future.

This PR adds a functionality to provide a predefined namespace which could be used to store a language-specific information. For example, the direction of the text in the language, or it's plural rules, or something else.

The user can use the default name for this namespace or change it to his/her own if it's necessary. 

Actually, there is nothing new. The user is able to store such info in the translation table without these functions. So, this is just like a shortcut, angular-translate-way. This also can be used to let the angular-translate know which namespace is used by the user to store language-specific information. So, we, probably, can use such information in some modules.

Advantages:
* unlike extending a translations() method of the provider, this solution allows the users to add language-specific information also in the json-files loaded from the server
* this method allows us to keep translation tables as a simple hash maps without extending them to more complex structures

Example for handling ltr/rtl languages:
```js
$translateProvider.translations('foo', {
  '___' : {
    'dir' : 'ltr'
  }, ...
});
$translateProvider.translations('bar', {
  '___' : {
    'dir' : 'rtl'
  }, ...
});
// ___ is a default name for predefined language-specific namespace
// some code
var dir = $translate.getLangInfo('dir');
if (dir === 'rtl') {
  // do something
} else {
 // do something else
}
```

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -25,6 +25,7 @@ angular.module('pascalprecht.translate').provider('$translate', ['$STORAGE_KEY',
       $loaderOptions,
       $notFoundIndicatorLeft,
       $notFoundIndicatorRight,
+      $langInfoNamespace = '___',
       NESTED_OBJECT_DELIMITER = '.';
 
   /**
@@ -444,6 +445,15 @@ angular.module('pascalprecht.translate').provider('$translate', ['$STORAGE_KEY',
     return this;
   };
 
+  this.langInfoNamespace = function(namespace) {
+    if (!angular.isString(namespace)) {
+      return $langInfoNamespace;
+    } else {
+      $langInfoNamespace = namespace;
+      return this;
+    }
+  };
+
   /**
    * @ngdoc function
    * @name pascalprecht.translate.$translateProvider#useMissingTranslationHandlerLog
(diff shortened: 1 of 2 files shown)


knalli:
I'll close this due inactivity.

Although this is perhaps a nice idea, there was no effort in the last months. If someone really things he/she needs this and/or will come up with a solution.. we are welcome this for sure.

DWand:
There is also an alternative solution for this thing. Something similar to old messageformat's approach of handling locales. Instead of adding the predefined namespace we can wrap this information in special services and add them to the `$translate`. Something like this:

```js
// add a new service
module.service('EnglishLanguageInformationProvider', [
function EnglishLanguageInformationProviderFn() {
  this.name = 'English';
  this.direction = 'ltr';
  this.getPluralCategory = function(n) { return (n === 1) ? 'one' : 'other'; };
}]);

// register this service
$translate.addLanguageInformation('en', 'EnglishLanguageInformationProvider');
```

This way we can avoid adding such information directly to translation tables. But also it will become more painful to add such information to the application (more services, more files, more configuration...).

PR #482, opened 2014-03-30
Add option to define a default translation text …

knalli:
This adds an optional opt-in feature:

``` html
<span translate=„my.key.for.a.welcome.message“ translate-default=„Welcome {{username}}!“></span>
```

or even

``` html
<span translate translate-default=„Welcome {{username}}!“>my.key.for.a.welcome.message</span>
```

### DIFF
--- a/src/directive/translate.js
+++ b/src/directive/translate.js
@@ -106,6 +106,10 @@ angular.module('pascalprecht.translate')
           }
         });
 
+        iAttr.$observe('translateDefault', function (value) {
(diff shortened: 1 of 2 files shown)


molerat619:
I'm sorry but after 20 minutes of reading issues, I finally want to ask, if  $translateProvider.useDefaultTranslationWhenNotFound() has been implemented and how I can use it.

I want to let a translation for a key fallback to english if the key is not available in the current language.

PR #496, opened 2014-04-04
function $translate.refresh() does not trigger "deferred.reject()" in case the translation part fails to load

npuzin:
then we cannot properly handle a loading failure.

This fix allows to write :

$translatePartialLoader.addPart('login');
$translate.refresh().then(function() {
  console.log('the login part has been loaded successfully');
}, function () {
  console.log('the login part faild to load');      
});


knalli:
Closing this right now due inactivity. Regarding the last concerns, a potential re-submit should be aware of the breaking/non-breaking situation.

PR #551, opened 2014-05-13
Instant interpolation of unkonwn

rewoo:
If a translation id does not exists, $translate.instant() now interpolates the
translation id. This supports gettext-ish translations where the
translation id is the base translation text.

Example:

		// Configuration
		$translateProvider.translations('en', {});

		// Code
		var result = $translate.instant('Hi {{ name }}!', {name: 'John Doe'});
		// interpolated result is 'Hi John Doe!'

I refactored Interpolator() and interpolateTranslation() for DRY. Hope that is fine.


knalli:
Closing due inactivity.

PR #590, opened 2014-06-04
Prevent calling then() on undefined for getTranslationTable() method.

Francisc:
`then()` is called on `undefined` when attempting to translate a non-existing key.


knalli:
@Francisc 

Could you please make a demo when this situation will happen? You can use this plnkr as a template: http://plnkr.co/edit/BJJwHaHnVFOIVYQjGuM2?p=preview

knalli:
Okay, we had already such a patch. Closing this.

PR #616, opened 2014-06-23
Added isPartLoaded and getRegisteredParts to $translatePartialLoader

etaroza:
I needed these changes to write a more robust translate-cloak. The problem is $translateLoadingSuccess is fired on every part load. Therefore current implementation of translate-cloak doesn't work correctly.

So I added a couple of methods and wrote my own cloaker.


knalli:
landed as 54f8ab3

PR #618, opened 2014-06-23
Fixes #566

### LINKED ISSUE #566: Provide API to list all languages that are configured by angular-translate
For implementing a language chooser, it would be nice to use ng-repeat to iterate over all available languages that are configured in angular translate. For this one need to have an api funcation that provides all configured languages.

Initiated through here: http://stackoverflow.com/questions/23659748/how-to-list-all-languages-that-are-loaded-by-angular-translate/

KevinLlopart:
getAvailableLanguageKeys() reuses internal table management logic, in order to avoid user mistakes and redundancies.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1583,6 +1583,20 @@ angular.module('pascalprecht.translate').provider('$translate', ['$STORAGE_KEY',
 
         return result;
       };
+      
+      /**
+       * @ngdoc function
+       * @name pascalprecht.translate.$translate#getAvailableLanguageKeys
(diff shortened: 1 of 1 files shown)


knalli:
I'm closing this now. In case of being urgent again, we should re-discuss this again.

Aides359:
@knalli @PascalPrecht May I take it that with [`return $availableLanguageKeys;`](https://github.com/angular-translate/angular-translate/blob/master/src/service/translate.js#L895) I could use (apart from feeling really dirty with calling _register_ as a getter) `registerAvailableLanguageKeys()` to get the registered languages?

PR #640, opened 2014-07-09
Load a fallback language only if it is not loaded already

sajithvk:
Load a fallback language only if it is not loaded already. If a lanaguage is already loaded through javaScript $translateProvider.translation(), we should not load it again.

TODO: Check if this condition need to be applied in other places as well.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1597,7 +1597,11 @@ angular.module('pascalprecht.translate').provider('$translate', ['$STORAGE_KEY',
         if ($fallbackLanguage && $fallbackLanguage.length) {
 
           for (var i = 0, len = $fallbackLanguage.length; i < len; i++) {
-            langPromises[$fallbackLanguage[i]] = loadAsync($fallbackLanguage[i]);
+            // Load a fallback language, only if it is not loaded already
+            //TODO: Check if this condition need to be applied in other places as well.
+            if($translationTable[$fallbackLanguage[i]]=== undefined){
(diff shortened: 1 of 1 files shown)


knalli:
Perhaps @tspaeth can take care of it, I mean, he had digged right into the fallback issues last week.

marcalj:
Any update? :)

knalli:
@tspaeth ?

tspaeth:
woops... will look asap. my mistake :bulb:

knalli:
ping :)

tspaeth:
Looks good to me, but I would rename the description in the code / PR a bit.
It's not about "loading", but more about initializing "in memory" language tables before setting up the loaders to retrieve the corresponding language tables.

@knalli have we ever allowed to add languages later at runtime (at least for fallbacks?)? If not, that must be checked if the flow would still allow that (as it is already noted in the TODO of this PR)

Sorry for the slight delay, folks :-)

knalli:
The original issue was translations already defined via `translateProvider.translations(key, map)` could not avoid a fallback request (it has ignore them and replaced it on load). That specific issue was fixed in 3be14df8 (#981) and will be released in the upcoming 2.7.0

Additionally, sometimes a request has been made multiple times (combination of preferred and fallback languages). Must be fixed also.

Hopefully, I'm not missing something this (2.7.0) version works as expected. http://plnkr.co/edit/ho934XZcfvwkQByVOKVJ?p=preview

I'll close this (in forward).

Issue #519, opened 2014-04-22
Multiple xhr (200) calls


juaniyyoo:
Hello,

When I'm defining the $translateProvider, it calls the url "/i18n/blocks.json" 3 times.
  once for preferredLanguage()
  once for fallbackLanguage()
  once for useUrlLoader()

The translations shouldn't be loaded only once with useUrlLoader() ?

Here is the code I used :

```
homeApp.config( function ($translateProvider) {
  $translateProvider.preferredLanguage('fr_FR');
  $translateProvider.fallbackLanguage('fr_FR');
  $translateProvider.useCookieStorage();
  $translateProvider.useUrlLoader('/i18n/blocks.json');
});
```

Being newbie to angular-translate, I may take the wrong way.

Angular : v1.2.16
Angular-Translate : v2.1.0

PR #662, opened 2014-08-04
fix(translatePartialLoader, translateStaticLoader, translateUrlLoader): ...

devonhumes:
This fixes angular-translate/angular-translate#519. The problem with the current is it only fixes one specific issue with multiple XHR requests. Namely, what happens if you have a preferred language and fallback language.

This fix handles all of the different async loaders, so that resources are only requested once. $translatePartialLoader#deletePart can still handle deleting data.

There is also a test to demonstrate an issue with the partial loader. When the partial loader is used and multiple parts are loaded, if the previously requested parts have not loaded yet, then multiple XHR request are made for each part.

This should also fix #640.

### DIFF
--- a/src/service/loader-partial.js
+++ b/src/service/loader-partial.js
@@ -22,6 +22,7 @@ angular.module('pascalprecht.translate')
     this.name = name;
     this.isActive = true;
     this.tables = {};
+    this.langPromises = {};
   }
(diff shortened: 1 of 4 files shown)


knalli:
Closing this due inactivity.

Issue #431, opened 2014-03-07
$translateProvider.determineLanguage does not work on Safari as expected


mohlendo:
When I use `$translateProvider.determineLanguage` in combination with `$translateProvider.registerAvailableLanguageKeys`, I would expect that the language key `en_US` would match case-insensitive, but it doesn't. 
The following plunker shows that bug. On Safari, no matching language key is found and you see no translation. On Chrome it works as expected.

http://plnkr.co/edit/gFzzNh?p=preview

knalli:
Closing this, we have the open PR (#755). That is still valid.

knalli:
The PR was replaced with a feature named uniformLanguageTag. Preview available [here](http://angular-translate.github.io/docs-canary/#/api/pascalprecht.translate.$translateProvider#methods_uniformlanguagetag).

PR #755, opened 2014-09-30
fix(service): reformat detected locales to something like "en_US"

yllieth:
fixes Issue #431


knalli:
Replaced with b534e1a

PR #768, opened 2014-10-06
fix(service): fallback languages follow shortcuts - fix #758

### LINKED ISSUE #758: Fallback language fails for missing shortcuts
You know this nice but not very useful feature for the lazy, Shortcuts. It lets you type `foo.bar` instead of `foo.bar.bar`.
It seems to mess with the fallback language functionality, when in the default language, `foo.bar.bar` happens to be missing.
The fix, at the moment, is to not use shortcuts, then it works as expected.
Please see the very simple plunkr here for demo: http://plnkr.co/edit/IsezYEIgcR1uTB2TKj7R?p=preview

tspaeth:


knalli:
replaced by #871

PR #853, opened 2014-12-10
Support for loading multiple static files

Zertz:
I added support for loading multiple files from different locations by passing a `files` key containing an array of objects.

I reused the previous code, wrapping it in a `load(file)` function. In order to remain backward compatible while keeping code simple, if only an object is passed, it is inserted into the `files` array. Then, it loops through every file and waits for all promises to resolve before merging the returned data into a single object.

Since files are requested asynchronously and then merged, duplicate keys will be overwritten in an unpredictable manner.

**Examples**

*Single file*

    $translateProvider
      .useStaticFilesLoader({
        prefix: '/path/to/locales/',
        suffix: '.json'
      });

*Multiple files*

    $translateProvider
      .useStaticFilesLoader({
        files: [{
          prefix: '/path/to/locales/',
          suffix: '.json'
        }, {
          prefix: 'relative/path/to/locales/',
          suffix: ''
        }, {
          prefix: '/another/path/to/locales/',
          suffix: '.json'
        }]
      });

Perhaps this can be improved, but the downside is that the minified script goes from 616 bytes 997 bytes, *a 60% increase*.

Questions? Comments?

### DIFF
--- a/docs/content/guide/en/12_asynchronous-loading.ngdoc
+++ b/docs/content/guide/en/12_asynchronous-loading.ngdoc
@@ -76,13 +76,32 @@ the following information is required:
 * **suffix** - specifies file suffix
 
 Let's say you have two localization files `locale-de.json` and  `locale-en.json`.
-You can simply use the staticFilesLoader with this pattern like this:
+You can simply use the loader like this:
 
 <pre>
 $translateProvider.useStaticFilesLoader({
     prefix: 'locale-',
     suffix: '.json'
 });
+$translateProvider.preferredLanguage('en');
+</pre>
+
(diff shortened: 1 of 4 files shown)


knalli:
@DWand maybe some spare time for a short translation change for ru/uk in `pr/853`?

knalli:
@DWand you can push directly in this branch, it is rebased on our current `canary`.

DWand:
@knalli, yes. I'll do it tomorrow.

knalli:
I've to rename the branch: `wip/pr853`

knalli:
Thanks all, landed as c462ee6, 8cd3456

PR #857, opened 2014-12-12
docs(guide): fix typo - missing parameter '$rootScope'

DWand:
Add missing parameter '$rootScope' in examples


knalli:
landed as 941cf02

PR #871, opened 2014-12-26
Extended (re)fix based on current canary - supersedes #768

tspaeth:
For issue description, see #758 

If everything looks good, the old #768 should be closed.


knalli:
landed as cce897a

PR #767, opened 2014-10-06
fix(directive, service): return value of translate-default also in case ...

tspaeth:
...fallback languages are used - related to #765

Issue #765, opened 2014-10-03
translate-default fails to work if using fallback languages


tspaeth:
As mentioned by @eekboom in https://github.com/angular-translate/angular-translate/pull/482 - the translate-default attribute fails to render if the module also has to iterate through the fallbacks.

The reason is, that the promise gets resolved with the key for the fallback if no translation value for that key is found instead of rejecting.

See https://github.com/angular-translate/angular-translate/blob/master/src/service/translate.js#L1143

... this leads to the if part in the directive for telling the translation to have worked.
see https://github.com/angular-translate/angular-translate/blob/master/src/directive/translate.js#L167

I'm currently not sure, if we can fix that issue without introducing another attribute or something else for communication that default is present from the directive -> service.
@knalli what do you think?

knalli:
anything open after fcd6b3e?

tspaeth:
Hm, at a first, quick crosscheck: I would say no, nothing left...  :-)

PR #877, opened 2015-01-10
fix(directive, service): return value of translate-default also in case ...

tspaeth:
...fallback languages are used - related to #765 (replaces #767)


knalli:
Thanks. fcd6b3e

PR #884, opened 2015-01-15
Fix default interpolation of {{ value + value }} with "escaped" sanitize value strategy

felixmosh:
Also added related tests for escaping the values


knalli:
@felixmosh  Thank you, landed as 75b7381

PR #893, opened 2015-01-22
feat($translatePartialLoader): adds optional priority param to the addPart function

dereksantos:
Allows parts to be loaded in order of their specified priority to avoid race conditions
between parts. Before this change, you could not guarantee the order in
which translation tables would be overridden.

This commit also fixes some grammatical errors in the `addPart` documentation.

### DIFF
--- a/src/service/loader-partial.js
+++ b/src/service/loader-partial.js
@@ -103,6 +103,15 @@ angular.module('pascalprecht.translate')
     return dst;
   }
 
+  function getPrioritizedParts() {
+    var prioritizedParts = [];
+    for(var part in parts) {
+      prioritizedParts.push(parts[part]);
+    }
+    prioritizedParts.sort(function(a, b){return a.priority-b.priority;});
+    return prioritizedParts;
+  }
+
 
   /**
    * @ngdoc function
@@ -278,11 +287,8 @@ angular.module('pascalprecht.translate')
         tables.push(table);
       }
 
-      var prioritizedParts = [];
-      for(var part in parts) {
-        prioritizedParts.push(parts[part]);
-      }
-      prioritizedParts.sort(function(a, b){return a.priority-b.priority;});
+
+      var prioritizedParts = getPrioritizedParts();
 
       angular.forEach(prioritizedParts, function(part, index) {
         if (part.isActive) {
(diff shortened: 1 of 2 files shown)


tspaeth:
From the code review perspective - LGTM. Well done ;-)

As I'm not a user of the part approach - I can't give an opinion on the functional usability.

dereksantos:
I've updated the code and test to make sure the priority is applied regardless of what order the http responses are received. Thanks for your feedback DWand.

dereksantos:
I'll remove the unnecessary lines.
It sounds like we should only be processing tables for parts that are active. In fact I think this is what the code used to do. So we need to add `if (part.isActive)` around that code.
I'm not very familiar with the use-cases for active vs. non-active parts, so let me know if that makes sense.

DWand:
@dereksantos, non-active parts are parts which were temporary deleted. They are not included into the resulting table at the moment. But in case they are added again, these translation tables will not be fetched from the server again because they already have translation data. It is like an internal cache from loaded parts. (https://github.com/angular-translate/angular-translate/blob/canary/src/service/loader-partial.js#L405)

As for `if (part.isActive)` there are at least 2 possible ways: add this condition in merging block or filter active parts in the `getPrioritizedParts` method.

dereksantos:
I'll go with the second option.

Also, I believe these lines can be removed since the else block can never be reached.  The `loaders.length` condition will always be true, and an empty array should still resolve with `{}`.
https://github.com/dereksantos/angular-translate/blob/prioritized-partial-loader/src/service/loader-partial.js#L304
https://github.com/dereksantos/angular-translate/blob/prioritized-partial-loader/src/service/loader-partial.js#L317-L319

DWand:
Hm... From docs it is not clear how the `$q.all` method will act with empty arrays.

dereksantos:
Here's a test plnkr. It appears to work as expected.  
http://plnkr.co/edit/8BUluatfWNb8AWAIYvgc?p=preview

DWand:
Okay, sources say that the `$q.all` method in such situation will immediately resolve the promise with an empty array (https://github.com/angular/angular.js/blob/master/src/ng/q.js#L532-534). So, it is okay.

DWand:
Looks good to me too. Thanks. +1 )

dereksantos:
That's good. You're right about 0 being falsy. In any case, I think it's a bit cleaner without that condition. 
Let me know if you'd like to see any other changes.

dereksantos:
Thanks for your feedback and help with this pull request!

knalli:
landed as 570617c

PR #910, opened 2015-02-04
fix(service): use $window/$windowProvider instead of window (fixes #856)

### LINKED ISSUE #856: Use $window instead of window
As Angular philosophy, `window` object should not be used directly. Use `$window` service instead.
[Example](https://github.com/angular-translate/angular-translate/blob/df3738119677b6834b2d47455bf35161c9b0c588/src/service/translate.js#L40)

tspaeth:
fix(docs): update ngdocs on storage-local.js

#856 fix


knalli:
wow.. so simple and it didn't worked for me :(

:+1:

knalli:
picked

PR #913, opened 2015-02-07
Improving the handling of interpolator id's that don't exist

liamkeaton:
Experienced an issue when loading the messageformat interpolator because I used the key messageFormat.

Perhaps it should error instead of falling back to the default interpolator?


knalli:
@liamkeaton I'm a little bit confused.. what is in your `interpolatorHashMap[interpolationId]` ? There should be nothing else then an interpolator...

liamkeaton:
@knalli I was getting undefined back when I was using the wrong interpolationId which broke interpolation and did not give me any useful errors.

knalli:
Oh, yeah. Got it. Just missed the fact it is an user input actually..

knalli:
Thank you, picked.

PR #922, opened 2015-02-11
Fix bower injection error with grunt-wiredep

tasos-ale:


knalli:
Thank you. Slightly modified, and picked.

PR #936, opened 2015-02-21
fix(loader): multiple static translation files

cleargif:
Two problems:

1. The docs has a typo,  `files` parameter is an object

2. The `url` arr is not looking up values from the passed `file` object


knalli:
Modified picked.

PR #937, opened 2015-02-23
docs(landingpage): fix messageformat link

marcin-wosinek:


knalli:
@marcin-wosinek  thank you very much, it's picked!

PR #948, opened 2015-03-06
fix(directive): Make interpolate message format work smoothly also on me...

### LINKED ISSUE #789: Errors when using translate directive with dynamic value and messageformat pluralization
Please see the plunker example:
http://embed.plnkr.co/RhuPW6cqpfsSuBS1R45a/preview

As far as I understand it's because ```angular-translate.js``` tries to submit ```translate-values``` while those are not ready. ```messageformat.js ``` library receives empty values and throws an error.

tspaeth:
...ssage format > 0.1.7 - fixes #789 

This is an experimental patch for the issues mentioned on messageformat stuff - e.g. in #789

### DIFF
--- a/src/directive/translate.js
+++ b/src/directive/translate.js
@@ -136,6 +136,16 @@ angular.module('pascalprecht.translate')
           }
 
           if (angular.equals(translationId , '') || !angular.isDefined(translationId)) {
+            // initially fetch all attributes if existing and fill the params
+            if (translateValueExist) {
+              for (var attr in tAttr) {
+                if (Object.prototype.hasOwnProperty.call(iAttr, attr) && attr.substr(0, 14) === 'translateValue' && attr !== 'translateValues') {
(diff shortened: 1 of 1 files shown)


knalli:
I think that is correct. Fine! :+1:

tspaeth:
Hm. Just figured out, that this doesn't help on translate-values - just translate-value... 

and translate-values seems a bit harder.

The way should be: Wait, until all attributes have been parsed, then run interpolation and messageformat...

tspaeth:
Didn't have time to update the PR yet... but for the impatient:
add 
```
       // initial setup
        if (iAttr.translateValues) {
          angular.extend(scope.interpolateParams, $parse(iAttr.translateValues)(scope.$parent));
        }
```

right here:
https://github.com/tspaeth/angular-translate/blob/canarycurrent/src/directive/translate.js#L127

knalli:
Picked as 2533f2d, 887dc1b

Issue #969, opened 2015-03-21
$translate.use should be able to fail/try again if failed once while loading a file


JamesBoon:
Hi,

if you are using the `$translateStaticFilesLoader` with `$translate.use(key)` and get a 404 once, you can never load the file again. The returned promise will always be resolved before the actual file is re-requested.
This affects all versions.

To fix it, `langPromises[key]` must be cleared after `$translate.use(key)` failed.
Adding `langPromises[key] = undefined;` before [emitting the change error](https://github.com/angular-translate/angular-translate/blob/canary/src/service/translate.js#L1543) should fix it.

Here is a test which will fail before the fix:
```javascript
var $translate, $httpBackend, $translateStaticFilesLoader;

beforeEach(module('pascalprecht.translate', function ($translateProvider) {
  $translateProvider.useStaticFilesLoader({prefix : 'lang_', suffix : '.json'});
}));

beforeEach(inject(function (_$translate_, _$httpBackend_, _$translateStaticFilesLoader_) {
  $httpBackend = _$httpBackend_;
  $translate = _$translate_;
  $translateStaticFilesLoader = _$translateStaticFilesLoader_;
  $httpBackend.when('GET', 'lang_nt_VD.json').respond(404);
}));

afterEach(function() {
  $httpBackend.verifyNoOutstandingExpectation();
  $httpBackend.verifyNoOutstandingRequest();
});    

it('try load the file again', function () {
  var wasResolved = false,
	wasRejected = false,
	promise = $translate.use('nt_VD');
  promise.then(resolved, rejected);
  $httpBackend.flush();

  expect(wasResolved).toEqual(false);
  expect(wasRejected).toEqual(true);

  promise = $translate.use('nt_VD');
  promise.then(resolved, rejected);
  $httpBackend.flush();

  expect(wasResolved).toEqual(false);
  expect(wasRejected).toEqual(true);

  function resolved() {
	wasResolved = true;
  }
  function rejected() {
	wasRejected = true;
  }
});
```

knalli:
fyi: I've made a proposal already in 9f1b0d1 related to #824

JamesBoon:
Hi @knalli,
thank you for the info.
I have created some tests for this and had some problems with this proposal. The PR #970 fixes this. Please have a look and tell me what you think.

JamesBoon:
Closed by #970

PR #970, opened 2015-03-22
fix(service): always remove stored ref for lang promises

### LINKED ISSUE #824: Second load from failed (404) static files load reads from internal tables instead of retrying the actual file load
http://plnkr.co/edit/309zz2RGQmtmDAoD8q21?p=preview

You load a file using static files loader, and if the file doesn't exist, it fails, gives a 404, and does not change the language (expected). However, if you load it again, angular-translate must have loaded the undefined result into internal translation tables, because pressing it for the second time will bypass the static files loader (can be confirmed by placing a console.error in that module's error function) and the keys will be loaded instead.

The desired behaviour should be that if there is a 404, the request to load a language should keep going through the static files loader until successfully loaded into angular-translate's internal table.

Cheers!

Daniel

JamesBoon:
If using a static file loader with $translate.use(key) an unavailable resource will no longer cause the promise to be successfully resolved if requested a second time.

Closes #824, #969

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1362,6 +1362,13 @@ angular.module('pascalprecht.translate').provider('$translate', ['$STORAGE_KEY',
         return result;
       };
 
+      var clearNextLangAndPromise = function(key) {
+        if ($nextLang === key) {
+          $nextLang = undefined;
+        }
+        langPromises[key] = undefined;
+      };
+
       /**
        * @ngdoc function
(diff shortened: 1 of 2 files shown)


knalli:
I've changed a little bit 

```js 
          langPromises[key]['finally'](function () {
            clearNextLangAndPromise(key);
          });
```


Thank you for contribution. Landed as dbd5be9

JamesBoon:
cool, ty :smile:

PR #974, opened 2015-03-24
docs(fix): remove old lang specific dependencies for newer messageformat...

tspaeth:
... versions in the docs section

fixes 404 on not finding locale de.js and en.js


knalli:
landed as fa05144

Issue #971, opened 2015-03-23
Got 'undefined' when using angular-translate directive


Uxspy:
Hi, we've upgrade angular-translate from 2.5.2 to 2.6.1 and found regression issues when using translate directives. Below is the html templates:

```html
<div translate>{{label}}</div>
````

When $scope.label is **undefined**, in angular-translate 2.5, the translated result is:

```html
<div></div>
```

However in angular-translate 2.6, the result is:

```html
<div>undefined</div>
```

Could anyone tell me is this a change or this is a bug? Thanks!

tspaeth:
@knalli due to the #854 fix on #701 - https://github.com/angular-translate/angular-translate/blob/master/src/directive/translate.js#L147-L150

On possible fix is to extend the updateTranslations to check for undefined... but does it have any other consequences?

```
if (translationIds.hasOwnProperty(key) && translationIds[key] !== undefined) {
```

knalli:
:shipit: :)

knalli:
Fix landed as 3f7cf4c

Uxspy:
Thanks, that's Great!

PR #975, opened 2015-03-24
fix(directive): handle interpolation of undefined keys correctly in upda...

tspaeth:
...teTranslations, fixes issues in #971


knalli:
Landed as 3f7cf4c

Issue #500, opened 2014-04-08
How to grab translation from another locale?


miraage:
For example, I've translations "en" and "ru".
I want to grab some translationId from "ru" locale while my active locale is "en".
Function getTranslationTable() is private, I cant reach it from my controllers.

What I need to do?

mdemille:
Has this had any progress? I'm in need of a way to get the translation table, just like the original poster. See my stackoverflow question: http://stackoverflow.com/questions/29240325/angular-translate-get-translation-table

knalli:
@mdemille nope, that is not planned right now. However, we like PRs... ;)

mdemille:
I think I got it. Let me know if I need to do anything else 

https://github.com/angular-translate/angular-translate/pull/977

miraage:
@mdemille, imho, pull should look like @knalli answer.

utkant:
I have a case where I would like the possibility to tell the translate tag which language to use.

Like so:

<p translate="keyHere" translate-language="ScopeVariableHere/LanguageKeyHere">

Hope this will be taken into consideration when creating this.

It would also be cool if we could set the default language for parts of the page like so:

<div translate-default-language="ScopeVariableHere/LanguageKeyHere">
All html elements inside this element will use the translate-default-language language...
</div>

skiant:
I've got a similar need for an app I'm working on (display a directive three times in a single view, each time in a different language), and I found this ticket while looking for a solution.

Is there something I can do or progress made already?

knalli:
No, speaking for myself this won't coming in 2.x anymore. However, it's already planned and implemented as draft in a demo-branch for v3.

skiant:
Thanks for the quick reply.
I guess the changes in v3 architecture would make this very hard to backport into v2, and v3 isn't ready for production use?

This feature is quite important for our users, hence I'll probably need to find a way to circumvent the current limitation.

knalli:
Yes, there are no backports planned. The current architecture does not allow a simple language change for a single request; it would be hardly to maintain.

skiant:
Got it, thank you very much for your time.

odedniv:
This is allowed on #1328 and added to 2.9 milestone. Should this be closed?

JerryBels:
@odedniv maybe wait for 2.9 to be released and then close ? :)

Great work btw !

PR #977, opened 2015-03-25
feat(service): add getTranslationTable method

mdemille:


knalli:
@tspaeth @DWand @PascalPrecht 

I would prefer a more elegant API supporting the whole stack w/ fallback, alias-resolving and so on. What do you think?

tspaeth:
@knalli yes. For example - based on what you've written here https://github.com/angular-translate/angular-translate/issues/500#issuecomment-39952291?

PR #990, opened 2015-04-08
feat($translatePartialLoader): accept function in urlTemplate

just-boris:
It allows to solve a lot of issues in compicated app structure.

For example I have a translation of common texts in my app: `/app/locales/en.json`.
And there is plugins which also have a translation `/plugins/{part}/en.json`.

Now I can't build so complex urls. But if I could pass a function inside, it would be possible


knalli:
@just-boris  thank you. I've adjusted a little bit, landed as 401204a

PR #996, opened 2015-04-10
[docs] "Paragraph" is "Absatz" in German

thp:


knalli:
@thp  Thank you, landed as 23684ef

PR #1004, opened 2015-04-13
docs(guide/uk,ru): Add link for a solution of the FOUC problem

DWand:
Adds a link to a solution of the FOUC problem to both Ukrainian and Russian version
of the Guide.


knalli:
Thanks. Picked

PR #1007, opened 2015-04-15
add support for update message at runtime

imbolo:
add support for update message at runtime


DWand:
It may be not so easy. There are also questions about events propagation, synchronizations with `use` and `refresh` methods and so on.

knalli:
Yep. We cannot expose the internal data map. This must be protected and handled, just like @DWand said already.

knalli:
We cannot expose the internals. Closing this.

Issue #988, opened 2015-04-07
$translate.refresh() doesn't clear all keys of $translateTable.


gsaini:
I am using angular-translate on my project. Where I loads the vocabulary on module basis. Now on change preference on application level, I need to clear all the previously loaded vocabularies. I am using $translate.refresh() to load the keys again on further navigation on modules but it loads the translates for current used key only while I want to load the translations keys again on navigate on that particular module after use $translate.refresh() once.

DWand:
Hello, @gsaini.

The purpose (or, probably, task) of the `$translate.refresh()` method is to refresh translation tables which are currently present in the `$translate` module. So, this method deletes old translation tables from the module and asynchronously loads their newer version.

If you want to *change* the translation table, then you have to combine the `$translate.refresh()` method with some loader. Something like this:

1. Configure the loader to return another translation table
2. Call the `$translate.refresh()` method

gsaini:
Can you please elaborate which you suggested above or any eaxmple?

& I suggest, It would be great if you can provide $translate.refreshAll() method which can be use to clear all translate tables.

DWand:
Hm... Really there is a problem. Thanks, @gsaini.

I'm not sure if I am right, but this scenario will lead to error:
1) Use some `lang1` language
2) Use some `lang2` language
3) Refresh translation tables
4) Use the `lang1` language

Here we will get the old version of the `lang1` translation table. It happens because we delete only those translation tables which were re-loaded from the server.

See here: https://github.com/angular-translate/angular-translate/blob/a2ac7000f94bcee6dd8e206cab5d7cec3527dd25/src/service/translate.js#L1651-L1654

@angular-translate/core, is it appropriate to delete **all** existing translation tables in case the new ones were loaded successfully? Do we need to re-load only the current languages or all languages which were loaded during the current session (this question can also affect separation of fallback language handling)?

knalli:
In general, a `refresh()` should refresh the things. So it is looking from the outside like being a bug when a translation has not been refreshed (case: a language is neither a fallback nor active but has been used in the past). I'm not sure if we really should *reload* all the languages. It would be sufficient ditching them only.

However, one problem is not solved and perhaps that's the original intent for that refresh strategy: If the reloading language was not injected via an async loader (i.e. predefined statically), the translations would be gone forever. This is actually true already nowadays, but only when being a fallback or in use.

DWand:
[...]

PR #1012, opened 2015-04-20
Fixes bugs with refresh method

DWand:
Fixes some tow bugs with the `$translate.refresh()` method:

1) Bad test for current table refreshing case
2) This method has to clear all translation tables if no language is specified (#988)

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1714,10 +1714,8 @@ function $translate($STORAGE_KEY, $windowProvider) {
           }
 
           var allTranslationsLoaded = function (tableData) {
+            $translationTable = {};
             angular.forEach(tableData, function (data) {
-              if ($translationTable[data.key]) {
-                delete $translationTable[data.key];
-              }
               translations(data.key, data.table);
             });
             if ($uses) {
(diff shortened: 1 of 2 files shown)


src/service/translate.js
```diff
@@ -1714,10 +1714,8 @@ function $translate($STORAGE_KEY, $windowProvider) {
           }
 
           var allTranslationsLoaded = function (tableData) {
+            $translationTable = {};
             angular.forEach(tableData, function (data) {
-              if ($translationTable[data.key]) {
-                delete $translationTable[data.key];
-              }
```
knalli: That's not a bugfix, just like noted [here](https://github.com/angular-translate/angular-translate/issues/988#issuecomment-94204364)

knalli:
Not saying it is wrong, but I would not say [this is a *bugfix*](https://github.com/angular-translate/angular-translate/pull/1012#discussion_r28755230).

Anyway. Have to look into the tests another day. Thanks!

src/service/translate.js
DWand: To keep things linked together.
I have just explained why I think that this is a bugfix [here](https://github.com/angular-translate/angular-translate/issues/988#issuecomment-94680684).

knalli:
As already said in #988. Fully ok. will pick this.

knalli:
Thank you @DWand

PR #1013, opened 2015-04-23
docs fix - "Flash of untranslated content" section

Dmitry9999:
If I understand the docs correctly, there is an important mistake in FOUC section - preferredLanguage is not the one which is loaded synchronously. Fixed it

### DIFF
--- a/docs/content/guide/de/12_asynchronous-loading.ngdoc
+++ b/docs/content/guide/de/12_asynchronous-loading.ngdoc
@@ -331,7 +331,7 @@ $translateProvider.useStaticFilesLoader({
     'prefix': 'locale-',
     'suffix': '.json'
 });
-$translateProvider.preferredLanguage('de');
+$translateProvider.preferredLanguage('en');
 </pre>
 
 **Hinweis:** Ein Benutzer von Angular Translate eine schöne Lösung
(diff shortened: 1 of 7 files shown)


Dmitry9999:
Should I reopen it to master branch instead of canary?

knalli:
@Dmitry9999 . You have made this correctly! The message style does not match the guideline, but I can take care for this.

knalli:
Well, I think the doc was actually already correct. The language codes must be different.

Here is an example: http://plnkr.co/edit/pZfWzSmoGskgd34KU0hK?p=preview

Dmitry9999:
@knalli Well, [the docs](https://angular-translate.github.io/docs/#/guide/12_asynchronous-loading) say
> you could provide at least one language with your app without having to load it asynchronously. If your app then **uses this provided language as default language**, using $translateProvider.preferredLanguage(), you won't have this FOUC, 

and then the code goes
`$translateProvider.translations('en', {`
and
`$translateProvider.preferredLanguage('de');`
where languages are different. In your plunker they are the same (both are 'default'). Am I missing something?

knalli:
If you define a different preferred language as the predefined (static) ones, you will have still an issue with FOUC.

Dmitry9999:
Exactly. This is what code in the docs does. preferred language and static language are different ('en' and 'de')

knalli:
@Dmitry9999  Ah, sorry! :(

You're right, totally. I've followed your version instead of the current one.

Thank you for fixing this! ;)

knalli:
Landed as af5d746

Thank you and sorry again. :)

Dmitry9999:
No problem :)

Issue #993, opened 2015-04-09
Translate directive should only render sanitized HTML when postCompile is off


marklagendijk:
The translate directive renders the text as HTML instead of text. Example:

``` javascript
{
  MY_LABEL: 'This is<br /> an example.'
}
```

``` html
<div translate="MY_LABEL"><div>
```
Is rendered as:
```
This is
an example.
```

This is a nice feature. However, when I found out about this feature I tested what would happen if I would put a script tag in the label. Turns out that it is rendered and the containing script executed. Example:

``` javascript
{
  MY_LABEL: 'This is<br /> an example.<script type="text/javascript">alert("Wait! What?");</script>'
}
```

This is a serious security issue. The solution would be to make it work in the same way as [ngBindHtml](https://docs.angularjs.org/api/ng/directive/ngBindHtml) by using the [$sanitize](https://docs.angularjs.org/api/ngSanitize/service/$sanitize) service. This will add a dependency on ngSanitize, but is the only way to do this properly.

knalli:
Valid point. That's why we have http://angular-translate.github.io/docs/#/guide/19_security -- original cause is backwards compatibility to AJS 1.2 (no ngSanitize).

marklagendijk:
Is it possible to configure it in such way that it uses $sanitize instead of doing `.text`? I do want to be able to use HTML in my labels, but don't want to introduce this security issue.

Also I don't agree with the current strategy of: let's keep this serious security issue for backwards compatibility. I would rather see a major version bump with this listed as a breaking change, and maybe the possibility to turn it off. I'm afraid that otherwise many people will overlook this in the documentation, like I did.

knalli:
> Also I don't agree with the current strategy 

Ehm, this was introduced while AJS 1.2 was available as stable.  There is no "strategy", it's simply not possible. Afaik there was no simple way to inject a non existing module.. ?

> I would rather see a major version bump with this listed as a breaking change, and maybe the possibility to turn it off. I'm afraid that otherwise many people will overlook this in the documentation, like I did.

Again, a valid point. Default security should be as high as possible (https://www.owasp.org/index.php/Establish_secure_defaults). A `3.0.0` will enable at least the `escaped` way (or even more) just like the documentation already said. Unfortunately, this project does need support by the community while we have a bunch of issues to do.

marklagendijk:
Oh, ok. So you are open to fixing this properly, which means introducing a breaking change? I could see if I can find the time to implement this.

[...]

PR #1014, opened 2015-04-23
feat(sanitization): refactored, fixed and extended sanitization #993

marklagendijk:
- Added $translateSanitization
- Removed sanitization logic from interpolators, and implemented usage of  $translateSanitization
- Changed $translate to use $translateSanitizationProvider for setting the strategy
- Added support for using multiple sanitization strategies, which will be executed as a chain.
- Throw an error when an unknown strategy name is specified. Failing silently would be a security issue.
- Added methods for adding / removing strategies to $translateSanitizationProvider.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -51,6 +51,7 @@ module.exports = function (grunt) {
         'src/service/translate.js',
         'src/service/default-interpolation.js',
         'src/service/storage-key.js',
+        'src/service/sanitization.js',
         'src/directive/translate.js',
         'src/directive/translate-cloak.js',
         'src/filter/translate.js'
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,8 @@
     "angular": "~1.2.26",
     "angular-translate-interpolation-messageformat": "2.6.1",
     "angular-mocks": "~1.2.26",
-    "angular-cookies": "~1.2.26"
+    "angular-cookies": "~1.2.26",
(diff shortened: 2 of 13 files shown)


knalli:
Thank you. Looks great. It would be great if you could apply some coding styles (`if (expr) {` or `} else {` like the existing ones. There are hint/lint errors, however, that's actually a regression made some months ago in the build chain I'd just figured out. Not your fault. ;(

Anyway.. first look promises a great feature. Everything is tested, even the case w/o the ngSanitize module. I'll have a depper look tomorrow I thing..

marklagendijk:
I adjusted the formatting settings in WebStorm and reformatted the code. Apart from that I fixed a small bug, and fixed the sanitization tests in the interpolators.

knalli:
@marklagendijk  Thank you very much. Landed as 12dbc57 with some fixups (additional info in log for changelog, and a few updates in the code)

marklagendijk:
@knalli you're welcome. Thanks for the way you helped with designing and reviewing this. I think we made an huge improvement to the security of angular-translate.

knalli:
Yes, you are absolutely right. That one has long been overdue.

marklagendijk:
@knalli question: when do you expect to be the new version (2.7.0?), with this feature in it to be released?

knalli:
@marklagendijk We have currently 1-2 open feedbacks in queue. After this, we are ready.

PR #1019, opened 2015-04-28
perf(directive): watch parameters only if exist

bekos:


knalli:
@bekos Thank you, landed as f0e2585

I've checked this multiple times, it still works. I'm still a little bit surprised we didn't have this already. Thank you again.

Issue #744, opened 2014-09-25
Add a prefix or nested translation


xnf:
Hi. Excuse me if this issue was discussed/resolved, but i burn in desperate desire to contribute. 
I use this wonderful module on a kind-of large system with thousands of translation keys and i feel a desperate need for some sort of grouping. 
Currently i prefix keys like SectionA:Button:Save, SectionB:Button:Save, but maybe you have thought about some general approach? For instance - adding a "translate-prefix/section/area" that would find same keys with: 
&lt;ANY translate-prefix="SectionA:" translate&gt;Button:Save&lt;ANY&gt;
&lt;ANY translate-prefix="SectionA:"&gt;{{'Button:Save'|translate}}&lt;/ANY&gt;
&lt;ANY translate-prefix="SectionA:" ng-bind="'Button:Save'|translate"&gt;&lt;/ANY&gt;

or perhaps i would be possible to use an object approach where values are objects like:
{"SectionA":{"Button":{"Save":"Save Section A"}}} and they can be adressed:
&lt;ANY translate&gt;SectionA.Button.Save&lt;/ANY&gt;
&lt;ANY&gt;{{'SectionA.Button.Save'|translate}}&lt;/ANY&gt;
&lt;ANY ng-bind="'SectionA.Button.Save'|translate"&gt;&lt;/ANY&gt;

Perhaps the later approach could go well with #604

irnc:
@knalli Besides the already implemented namespaces, this issue asks about another concept, namely the `translate-prefix` directive.

Consider following markup and the long namespaces used:

```html
<div class="welcome">
  <section id="login">
    <h1 translate>start-page.welcome.login.heading</h1>
    <p translate>start-page.welcome.login.help-text</p>
  </section>
  <section id="sign-up">
    <h1 translate>start-page.welcome.sign-up.heading</h1>
    <p translate>another.namespace.id</p>
  </section>
</div>
```

It would be more convenient to have `translate-prefix` directive at parent element, e.g.

```html
<div class="welcome" translate-prefix="start-page.welcome">
  <section id="login" translate-prefix=".login">
    <h1 translate>.heading</h1>
    <p translate>.help-text</p>
  </section>
  <section id="sign-up" translate-prefix=".sign-up">
    <h1 translate>.heading</h1>
    <p translate>another.namespace.id</p>
  </section>
</div>
```

What do you think about this approach? Would you be willing to see relevant pull request?

knalli:
[...]

Issue #641, opened 2014-07-10
Support for "namespaced elements"


spencerwi:
First of all, great lib. I like how simple it is to use.

The one thing that does bug me a bit is when I have big/deep pages, having to use constantly retype long nested-namespace translation ids like this whenever I want to properly scope my translations using namespaces:

    DASHBOARD.TABS.CHEFREQUESTS.SEARCHMODAL.USETHISITEM

It'd be nicer if I could use a directive on an element to "pre-build" parts of that tree; so I'd have have something like this:

    <div translate-namespace="DASHBOARD">
        ....
        <div translate-namespace="TABS">
            ....
             <div translate-namespace="CHEFREQUESTS">
             ....
                 <div translate-namespace="SEARCHMODAL">
                     ....
                     <a class="btn" translate>USETHISITEM</a>
                     ....
                 </div>
                 ....
             </div>
             ....
        </div>
        ....
    </div>

That way, I don't have to constantly repeat giant long nested namespaces, but I can still namespace my translations to keep them scoped to specific uses.

bcldvd:
+1

knalli:
Landed already in canary.

PR #1021, opened 2015-04-29
feat(directive): translate-namespace directive

tsubik:
I've added translate-namespace directive to support "namespaced translation keys". In order to use  given translation namespace translation key should be prefixed by "."

E.g. to get translation for key "client.posts.title"

```
<div translate-namespace="client.posts">
  <h1 translate>.title</h1>
</div>
```
or you can also nest translate namespaces
```
<div translate-namespace="client">
  <div translate-namespace=".posts">
    <h1 translate>.title</h1>
  </div>
</div>
```
Relates #641, #744

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -57,6 +57,7 @@ module.exports = function (grunt) {
         'src/service/storage-key.js',
         'src/directive/translate.js',
         'src/directive/translate-cloak.js',
+        'src/directive/translate-namespace.js',
         'src/filter/translate.js'
       ],
 
--- a/demo/ex13_translate-namespace.htm
+++ b/demo/ex13_translate-namespace.htm
@@ -1,0 +1,122 @@
+<!DOCTYPE html>
+<html ng-app='app'>
+
+<head>
+  <meta charset="utf-8">
+  <title translate="TITLE">Translate namespace directive</title>
+  <style>body { text-align: center; }</style>
+</head>
+
+<body ng-controller="ctrl">
+
+  <p>
+    <a href="#" ng-click="setLang('en_US')">English</a>
+    |
+    <a href="#" ng-click="setLang('ru_RU')">Русский</a>
+  </p>
+
+  <div translate-namespace="ns1">
+    <h1 translate>.HEADER</h1>
+    <h2 translate>.SUBHEADER</h2>
+  </div>
+
+  <hr>
+
+  <div translate-namespace="ns2">
+    <h1 translate>.HEADER</h1>
+    <h2 translate>.SUBHEADER</h2>
+  </div>
+
+  <div translate-namespace="ns3">
+    <div translate-namespace=".HEADERS">
+      <h1 translate>
+        .HELLO
+      </h1>
+    </div>
+    <div>
+      <h1 translate>ns4.HEADERS.HELLO</h1>
+    </div>
+  </div>
+
+  <div translate-namespace="ns5">
+    <h1 translate>.HEADERS.HELLO</h1>
+  </div>
+
+<script src="../bower_components/angular/angular.js"></script>
+<script src="../bower_components/angular-cookies/angular-cookies.js"></script>
+<script src="../dist/angular-translate.js"></script>
+<script src="../dist/angular-translate-storage-cookie/angular-translate-storage-cookie.js"></script>
+
+<script>
+angular.module('app', ['pascalprecht.translate', 'ngCookies'])
+
+.config(['$translateProvider', function($translateProvider){
+
+  // Adding a translation table for the English language
+  $translateProvider.translations('en_US', {
+    "TITLE"     : "How to use translate namespace directive",
+    "ns1" : {
+      "HEADER"    : "A translations table supports namespaces.",
+      "SUBHEADER" : "So you can to structurize your translation table well."
+    },
+    "ns2" : {
+      "HEADER"    : "Do you want to have a structured translations table?",
+      "SUBHEADER" : "You can to use namespaces now."
+    },
+    "ns3": {
+      "HEADERS": {
+        "HELLO": "Hello"
+      }
+    },
+    "ns4": {
+      "HEADERS": {
+        "HELLO": "Hello from ns4"
+      }
+    },
+    "ns5.HEADERS.HELLO": "Also works for simple keys with dots"
+  });
+
+  // Adding a translation table for the Russian language
+  $translateProvider.translations('ru_RU', {
+    "TITLE"     : "Как использовать директиву перевести имен",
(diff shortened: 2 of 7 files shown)


DWand:
Hello, @tsubik. Thank you for this PR.
Are there any reasons why this functionality should change the `$translate` service instead of generating a translation key inside the directive?

tsubik:
Good point. I had this code in translate directive at first and now I don't remember why I've moved it into service. I will move this functionality into `updateTranslation` function in directive.

marklagendijk:
@tsubik I think that the preceding `.` should not be needed. So:

``` html
<div translate-namespace="client">
  <div translate-namespace="posts">
    <h1 translate>title</h1>
  </div>
</div>
```

tsubik:
@marklagendijk Then you won't have any possibility to use translation key from out of the namespace.

marklagendijk:
@tsubik not necessarily. I would expect it to look in the current specified namespace, and if not found, go up the namespace tree to look if it is specified on that level.

``` javascript
var translations = {
  buttons: {
    ok: 'OK!'
  }
  myApp: {
    myForm: {
      buttons: {
        goAway: 'Go away!!'
      }
    }
  }
}
```

``` html
<div translate-namespace="myApp.myForm">
  <button translate='buttons.goAway'></button>
  <button translate='buttons.ok'></button>
</div>
```

knalli:
Hi @tsubik, thank you for contribution. Your PR is looking great and formally everything is okay (small note: have a look for spacing) . Brilliant!

I'm with @marklagendijk in this two points:

1. The preceding dot is based on a technically decision marking usage of namespacing. 
  1. In `translate-namespace` it can be removed definitely.
  2. In `translate` it depends:  it marks whether being namespacing will be used or not.
2. At least the namespace would override the current behavior. Given the last example by @marklagendijk, `buttons.ok` should work because it the _regular way_. Otherwise you will be forced with issues forgetting one small `translate-namespace` in path.
  1. I'm still not sure whether "going up" would not make the whole thing very complicate. Using again the last example: what if `myApp.buttons.ok` exist (really following each of the items of the namespace)?

I thing without making the things too much complicate (proposed):

```json
{
  "buttons": {
    "one": "A_ONE"
    "two": "A_TWO",
    "three": "A_THREE"
  },
  "app": {
    "buttons": {
      "one": "B_ONE"
      "two": "B_TWO"
    },
    "module": {
      "buttons": {
        "one": "C_ONE"
      }
    }
  }
};
```

```html
<body translate-namespace="app">
  <ol translate-namespace="module">
    <li translate>one</li>
    <li translate>two</li>
    <li translate>three</li>
  </ol>
</body>
```

would result into

```html
<body >
  <ol >
    <li>C_ONE</li> <!-- overriding the global -->
    <li>A_TWO</li> <!-- only B_TWO if we following all possible namespace items -->
    <li>A_THREE</li> <!-- translation rejected because of missing, so using global -->
  </ol>
</body>
```

So, now I've written this down I'm remember we have fallback languages, too.

Ah, and it should be noted this cannot work in filters (it is not covered here). Because if we add support for this (however, this will require an explicit scope like `{{ 'buttons.ok' | translate: this }}`) it would conflict with the opt-out of `$translateProvider.statefulFilter`.

DWand:
IMHO, `.button.ok` syntax looks more deterministic, it is much easier to understand what translation has to be performed in every particular case. So, it is better for debugging and maintaining.

tsubik:
I didn't want to overcomplicate this. Fallbacks are great but only on language level. I think that looking up the tree to find key would bring too much confusion to every project. It would be more difficult to locate which translateKey is used. If I had some defaults translations in project I would rather use `@:shortcut`.

```
var translations = {
  buttons: {
    ok: 'OK!'
  },
  myApp: {
    myForm: {
      buttons: {
        ok: '@:buttons.ok',
        goAway: 'Go away!!'
      }
    }
  }
}
```

And as @DWand has mentioned dot is used to notice that the key is located in some namespace. About the filter, We can work around this by using simple filter for prefixing key like so `{{ '.buttons.ok' | prepend:translateNamespace | translate }}`.

knalli:
Valid points.

```html
<body translate-namespace="app">
  <ol translate-namespace="module">
    <li translate>.one</li>
    <li translate=".two"></li>
  </ol>
</body>
```

We can skip the namespace's dot, but stay with the id's one. Ok? Then this is a directive only feature.

tsubik:
Hmm I am not sure. Let's say we have some complicated form which we want to reuse in many parts/modules in our application. We can have translations as follows:

```
var translations = {
  moduleA: {
  },
  moduleB: {
  },
  shared: {
    form: {
    }
  }
}
```
Then you set the namespace in your component to `shared.form` and if your modules have also set a namespaces then including this component in moduleA or B will not work, translations will be searched for in namespace `moduleA.shared.form` instead of in `shared.form`. I believe that in most cases nesting won't be needed and overwriting current namespace is more important. My proposition is to leave dots if you want to append to current namespace and without dot if you want overwrite. This approach will be more consistent. Anyway I need to change code to look current namespace also in parent scopes, right now it is not working with isolated scopes.

knalli:
Well, keeping it simple, there would be no such alternative strategies.

> Anyway I need to change code to look current namespace also in parent scopes, right now it is not working with isolated scopes.

I've still not checked out the source (running, testing). But after only looking into it, I guess there is also properly an issue with parallel scopes. I think the namespace should not be "computed" when configuring the namespace (in namespace-directive), but when building the translation request (in translate-directive)

tsubik:
It should work with isolated scopes now. What do you mean by parallel scopes?

knalli:
I've skipped this one for 2.7.1.. have not time for a deep usage review. But that's coming in next!

knalli:
Landed as 45523bb

knalli:
Thank you again!

jmls:
great work - I was looking for this exact functionality. How can I use the canary branch with bower ? I've tried a bower install with "canary" but got this error

bower angular-translate#canary     ENORESTARGET Tag/branch canary does not exist

Additional error details:
Available tags: 0.4.2, 0.4.3, 0.4.4, 0.5.1, 0.5.2, 0.6.0, 0.7.0, 0.8.0, 0.8.1, 0.9.0, 0.9.1, 0.9.3, 0.9.4, 1.0.0, 1.0.1, 1.0.2, 1.1.0, 1.1.1, 2.0.0, 2.0.1, 2.1.0, 2.2.0, 2.3.0, 2.4.0, 2.4.1, 2.4.2, 2.5.0, 2.5.2, 2.6.0, 2.6.1, 2.7.0, 2.7.1, 2.7.2
Available branches: master

thanks

PR #1029, opened 2015-05-06
this.replace is not a function

benjaminpick:
Hi,

when giving a non-string Id to the translate Filter, I am getting the error `TypeError: this.replace is not a function. Attached fix works for me.

(Use Case: I am using translation in tables that can contain labels as well as data.)


benjaminpick:
Sorry, I should have made a pull request against canary, closing this.

PR #1030, opened 2015-05-06
this.replace is not a function

benjaminpick:
Hi,

when giving a non-string Id to the translate Filter, I am getting the error `TypeError: this.replace is not a function. Attached fix works for me.

(Use Case: I am using translation in tables that can contain labels as well as data.)

### DIFF
--- a/src/directive/translate.js
+++ b/src/directive/translate.js
@@ -100,7 +100,7 @@ function translateDirective($translate, $q, $interpolate, $compile, $parse, $roo
    * @returns {string} The string stripped of whitespace from both ends
    */
   var trim = function() {
-    return this.replace(/^\s+|\s+$/g, '');
+    return this.toString().replace(/^\s+|\s+$/g, '');
   };
 
   return {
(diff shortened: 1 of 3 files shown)


tspaeth:
Thanks for your contribution. Even though the check might make sense in general, I'm asking myself:

You're not translating values/data here, right? You're just expecting a non-string value to be returned as-is?!

benjaminpick:
Well ... you're right, the flaw is deeper in my application logic. However, it took some time for me to find out the reason for this, as angular doesn't give line numbers when intrapolating statements. So I guess an error would be fine if it would be more clearly angular-translate-related ...

For completenes, the tests should also do integer, undefined and so on. I will add them on Friday.

benjaminpick:
I have added the current return values in the commented part. Esp $translate([]) == {} is a bit inconsistent but you're right, this normally shouldn't happen.

knalli:
Thanks. Landed as 8616dca

Issue #995, opened 2015-04-09
angular-translate 3.0.0


knalli:
We should talk about this :)

In general, this is not a new thing, however, it was only spoken and written down anywhere. We should plan a next major release, perhaps not with many new features. But with the opportunity to cut down some "legacy strings".

* [ ] Dropping support for AJS 1.2, updating internal base to AJS 1.3.x latest
  * [ ] `filter.$stateful` management is not optional anymore (if-branch can be omitted)
  * [ ] with a re-arrangement of the escaping strategies we can use `ngSanitize`
* [ ] better security by default (latest mention in #993), already announced in [docs](http://angular-translate.github.io/docs/#/guide/19_security))
* [ ] re-arrangement of modules
 * [ ] discussion state of satellite bower shadow repositories -- we should discuss whether we need them or we can combine them all files into one (dedicated) repository (reasons pro/contra?). 
 * [ ] re-think the current internal module dependency hierarchy -- do we need dedicated modules actually? Why? We do not need them for a plugin architecture.
 * [ ] a full working migration guide (more or less a well written list of the breaking changes)

This is community driven, so we need everyone's input, additions or comments.

/cc @angular-translate/core @angular-translate/members

DWand:
@knalli, and how about synchronization of processes which are taking place in `use` and `refresh` methods? It looks like a big fundamental piece of work which affects internal architecture of the service.

knalli:
What do you mean in detail?

DWand:
@knalli, I'm not sure if this issue is still valid, but most likely it is. I've described it in the issue #449. As I remember, there were also a few other issues somehow related to the question of synchronization.

DWand:
I disagree with at least 2 options:

1) I think that it is not good to make **all** things asynchronous. Why should something pretend to be asynchronous if it is synchronous by nature? I think that such tools have to be used only if they are needed to accomplish the task. Needless promises may confuse the developer and lead to misunderstanding of how the thing works.

2) The `activeLanguage` method looks like a getter. If we try to make things more clear and easy to understand, then it might be better to split the `use` method to `getActiveLanguage` and `setActiveLanguage`. This is Java style, but people use it for years and it works pretty well.

knalli:
1.  More the promise style allowing a deferred setup. And if A is internally deferred, than B will it be also automatically. 
2.  My idea was to stick with the current way of a chaining configuration.

knalli:
Addition notes on "re-arrangement of modules"

[...]

PR #1038, opened 2015-05-13
The $cookieStore service is deprecated.

mrzepinski:
Note: The $cookieStore service is deprecated. Please use the $cookies service instead.

https://docs.angularjs.org/api/ngCookies/service/$cookieStore

### DIFF
--- a/src/service/storage-cookie.js
+++ b/src/service/storage-cookie.js
@@ -3,7 +3,7 @@ angular.module('pascalprecht.translate')
 /**
  * @ngdoc object
  * @name pascalprecht.translate.$translateCookieStorage
- * @requires $cookieStore
+ * @requires $cookies
  *
  * @description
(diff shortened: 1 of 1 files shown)


knalli:
Well, it is marked as being deprecated in the upcoming and still unreleased AJS 1.4. 

We should evaluate the differences between both, $cookies is in AJS 1.2 already.

knalli:
@mrzepinski , thanks for the contribution. But this one is rejected. 

At first, the change removes support for both AJS 1.2 and AJS 1.3 (!). Technically, because the internal interface of `$cookies` has been changed from a simple POJO (with properties) to a object with getter/setter.

Secondly, `$cookieStore` is still valid, but would be probably ditched in AJS 1.5 (although that would not be semantically, but hey, that's AJS' yolo :))

The test fails because the internal change of `$cookies` was not handled.

I'll create a TODO (conflicting with the rewrite #995).

PR #1045, opened 2015-05-26
fix(docs): fix typo in $translateChangeSuccess

kvetis:
fix typo in event name in EN FR RU UK docs
- if incorrectly copied the code did not work


knalli:
Thanks, landed as 89e2569

Issue #1056, opened 2015-05-29
$translationCache Unknown provider


Nighthawk14:
Hi,

I'm trying to set up the translation cache to be used in complement of $templateCache, because right now If I prepare all the templates during my production build, the translations are not working. 

When I load the `$translationCache` like that :
```js
.run(function($translationCache) {
    console.log($translationCache);
  });
```

Angular return me an `Unknown provider` Error. I saw that the service :
`/src/service/translationCache.js` is not loaded in Grunt, so I guess it's missing during the build of the library and so the `angular-translate.js` file in the `bower-angular-translate` repository doesn't have the `$translationCache` factory.

And by the way, do you have an idea of how to get the translations working with the $templateCache, I can't find any workaround or examples :).

ps : I use the `StaticFilesLoader` to load my translations files

Thanks :),
Antoine.

tspaeth:
Yup, looks like you're right that during https://github.com/angular-translate/angular-translate/commit/b68560160ea51b07030b0cc0ebc4ece440546ad0 - the Gruntfile.js didn't get changed / that file added.

@knalli can you have a look?

knalli:
Fix landed as 2fcbc60, in upcoming 2.7.1

PR #1057, opened 2015-05-30
fix(service): integrate translationCache into service distribution file

tspaeth:
The missing piece fix for #1056


knalli:
I wonder why we have added this feature.. looks to be a "less than 5% feature". :S

knalli:
Landed as 2fcbc60

PR #1053, opened 2015-05-28
feat($translateProvider): add a new option to force async reload

acourtiol:
Linked to issue #1051.

PR #1059, opened 2015-06-01
docs(loader): German docs addition to PR #1053

tspaeth:


knalli:
Landed in bdee77f (squashed)

PR #1070, opened 2015-06-08
Abort the last loader if not finished

yannickglt:
In some cases, two requests are sent to get different i18n files. Sometimes, the second request finishes before the first one (because the file is weightless for example) and then the last loaded language is not the expected one. 
To avoid this behavior, I interrupt the previous request (if it is currently pending) before executing the new request.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -965,7 +965,8 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
           interpolatorHashMap = {},
           langPromises = {},
           fallbackIndex,
-          startFallbackIteration;
+          startFallbackIteration,
+          abortLoader;
 
       var $translate = function (translationId, interpolateParams, interpolationId, defaultTranslationText) {
 
@@ -1138,10 +1139,16 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
           cache = $injector.get(cache);
         }
 
+        if (pendingLoader && abortLoader) {
+          abortLoader.resolve();
+        }
+        abortLoader = $q.defer();
+        
         var loaderOptions = angular.extend({}, $loaderOptions, {
           key: key,
           $http: angular.extend({}, {
-            cache: cache
+            cache: cache,
+            timeout: abortLoader.promise
           }, $loaderOptions.$http)
         });
 
--- a/test/unit/service/translate.spec.js
+++ b/test/unit/service/translate.spec.js
@@ -534,6 +534,164 @@ describe('pascalprecht.translate', function () {
   });
 
   // Spec for edge case: while preferred language is still loading,
+  //                     another language will be requested and
+  //                     is returning before the preferred one
+  // t0 ~ 0.0s           (request ab_CD)
+  // t1 ~ 0.5s           (request en_US)
+  // t2 ~ 1.5s           (response en_US)
+  // t3 ~ 2.0s           (response ab_CD)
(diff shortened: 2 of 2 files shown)


yannickglt:
Hi guys!
I proposed this PR to solve a bug we got on production. Do you think the solution I proposed is good?
If it is, could you accept the PR or how could I improve it to be acceptable?

knalli:
Due lack of time, this one has not got be reviewed.

Actually, we should always have covered the correct order of requests. We must have a deeper look into it.

yannickglt:
@knalli Do you think there is a better way to do that? You can tell me what do you expect and I'll try to make the changes.

knalli:
We have some tests regarding order of requests (slow/fast loaders in parallel).

knalli:
https://github.com/angular-translate/angular-translate/blob/92bc9551279fc2f5bc7daaedebc87adede8b3713/test/unit/service/translate.spec.js#L443-L620 in particular

knalli:
Meanwhile: Commit message style, PR origin and target branch are not correct regarding our contribution guideline. I think we (me) can handle this on the small (meta) changes, but you should take care of it next time.

note to ourselves: The review must take the tests into account, perhaps with an additional test.

knalli:
@yannickglt  You you please add a test covering YOUR intentional purpose doing this PR?

I just want to be sure this will work both for you and for us as expected. Ideally add the tests just right after the other "timing tests" (see here https://github.com/angular-translate/angular-translate/pull/1070#issuecomment-131618200).

yannickglt:
Sure! I'll do it asap. Thanks for your advice.

yannickglt:
That's done!

knalli:
Well done, thank you very much. Landed as dd4a8b4

daniel-franz:
This cancels the request for the fallback language in our case using the static files loader.
This is our configuration:

```
    $translateProvider.useStaticFilesLoader({
        prefix: 'locales/',
        suffix: '/translation.json'
    });

    $translateProvider
        .registerAvailableLanguageKeys(['de', 'en'], {
            'de*': 'de',
            'en*': 'en'
        })
        .determinePreferredLanguage()
        .fallbackLanguage('en');
```

In version 2.7.2 and before translations not available in "de" were read from "en". As the request to "en" gets canceled now, only items from "de" are available. Was that intended?

yannickglt:
No I guess it is not @daniel-franz. I will investigate on this issue and let you know.

eyoung1980:
I'm having the same issue as @daniel-franz.

I was loading the app with the browser language set as fr and then dynamically changing the language to en-GB but because the fallback language request had been cancelled the translations were not available.

In my case removing the fallbackLanguage and setting a default in registerAvailableLanguageKeys solved my issue.
```
$translateProvider.registerAvailableLanguageKeys(["en-GB", "fr-FR"], {
            "en*": "en-GB",
            "fr*": "fr-FR",
            "*": "en-GB",
        });
```

yannickglt:
@knalli the PR #1225 should fix the issue raised by @daniel-franz and @eyoung1980.

I removed the abort solution as it does not concern only the `translate#use` method, but also `translate#fallbackLanguage` (and possibly every method calling `translate#loadAsync`). I chose another solution targeting only the `translate#use` method. The goal of #1070 was to avoid that the slowness of some requests modify the order of the languages to use.

A new unit test covers the case described above. You can try running it on your master branch (which should fail because of #1070) and on #1225 to check that the PR does the trick.

knalli:
Thank you for the patch. Sorry for the delay.

PR #1076, opened 2015-06-11
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


knalli:
Thank you, but we will stick with the version. The spec only describes it will be ignored _at the moment_.

PR #1078, opened 2015-06-11
Include example on wildcards in registerAvailableLanguageKeys

atd:
Fixes angular-translate/angular-translate.github.io#9


atd:
Ops! not canary

PR #1079, opened 2015-06-11
Include example on wildcards in registerAvailableLanguageKeys

atd:
Fixes angular-translate/angular-translate.github.io#9


atd:
I searched for the right word :smile: 

That is how it is called in Wikipedia http://en.wikipedia.org/wiki/Locale

But RFC 5646 uses *region* http://tools.ietf.org/html/rfc5646#section-2.2.4

tspaeth:
ISO 639 is more talking about "languages" and "macro languages"... mh.... so many options :-) http://en.wikipedia.org/wiki/ISO_639-3

region is also ok to me.

knalli:
country + territory is correct in that term.

knalli:
Thank you, landed as aeff411

PR #1141, opened 2015-07-22
Add support for query string params in useStaticFilesLoader options

kojiwakayama:


knalli:
@kojiwakayama  Thanks for contribution. However, why are using the *static files* loader when using a dynamic request? Would not be the url loader the better one?

kojiwakayama:
Thanks @knalli. Its possible to use the url loader but it would be a hack. Our use case is to cache bust the json files via query string params.

knalli:
Well, that means you are not wanting introducing params actually (which some kind of invalid in context of static files!), but using a custom `$http` option.

The reason why I'm rejecting this is clear: This loader should be easy, clear and simple to use. An option for dynamic params in context of static files is not anything of these.

If you need to add dynamic params, then you should use the appropriate loader (url loader).
If no built-in loader match your need, please use a custom one.

Note: It depends on the other parts of your system, but proper http header should work also. You can apply additonal $http options with `options.$http`.

See also more about this in this [meta AngularJS issue](https://github.com/angular/angular.js/issues/1418).

dylansmith:
@knalli there is no technical reason why query params are invalid in the context of any URL, whether it references static files or not. Cache busting is a perfectly valid and widely-used use case.

I personally don't see how adding an optional params object detracts in any meaningful way from the simplicity and ease-of-use of this loader; the primary use case is not impacted in any way. Forcing users to use `$translateUrlLoader` and provide semantically invalid values for the `queryParameter` and `key` options is less clear and straightforward in my opinion.

PR #1161, opened 2015-08-13
Added support for @: links within a translation string

aupp:
Added support for @: links within a translation string, like this "Bla bla bla @:foo.bar.text bla bla bla"


knalli:
Hi @aupp , thank you for contribution.

Regarding the contribution guidelines, we have some issues:

* the changes are split in separated commits, without need
* commit message style intend no useful changelog pattern
* a pull request should branched by the canary and to the canary

And technically, I wonder why the change was only made for `determineTranslationInstant` (meaning filter), but not `determineTranslation` (directive + service).

aupp:
@knalli My simple reply is because I'm a noob :) I never did a pull request before.
It slipped my mind adding the feature to the directive + service as well - I'll do that.
Regarding the commit messages - what should they look like?
And there are two commits because I realized I forgot to handle the special characters a while after my initial commit.

knalli:
@aupp Wonderful you are doing contributions :) We have prepared a document for this: https://github.com/angular-translate/angular-translate/blob/master/CONTRIBUTING.md

aupp:
So how do I do this. Should I remove this PR and do a new one with only 1 commit and the correct commit message style?

knalli:
@aupp Because you do not have targeted the correct branch on PR creation: yes, unfortunately.

You just have to:

* rebase your changes onto the latest changes of `canary` (be sure that you branch of `canary`!)
* merge/squash the commits into one
* use a proper commit message

and then make a new PR to OUR `canary`. That's all.

(The new PR is only required due GitHub limitation of not changing this any more. Rebasing would be possible at any time.)

knalli:
@aupp Do you still want to complete this PR?

aupp:
Oh, yes! I would. But I need to do it right this time. Had alot on my hands lately and forgot about this! Will try to do a proper PR asap!

knalli:
Hello,

I'm closing this issue due inactivity. We have asked for additional details, but you could not helped us. Therefore we close this issue as not reproducible.

If you still have an issue, please create a new issue because we cannot be sure to track closed issues.

PR #1165, opened 2015-08-17
$translateProvider.preferredLanguage() allows to get the preferredLanguage during configuration phase

michaellieberherrr:


jtomaszewski:
Yeah. It was working like this but then it has been broken in some version.

knalli:
Thank you, landed as 6014a81

PR #1171, opened 2015-08-19
feat($translateProvider): add option to customize nested delimiter

ca136:


knalli:
Thank you, landed as 6ddb102.

I've added an actual test for working ;)

PR #1181, opened 2015-08-27
Added lang_only as new langTagResolver

ThinklabsGbR:
Added possibility to be able to get only the lang tag as result, in order to limit the language files to one for each language, i.e. locale-de.json, locale-en.json. Especially useful when using the staticFilesLoader.


knalli:
@ThinklabsGbR 

Thank you for contribution. Please follow https://github.com/angular-translate/angular-translate/blob/master/CONTRIBUTING.md which contains some notes about contribution steps in this project.

Particular in this case, I remark the followings:

* apply a proper commit message (see in guideline about conventional changelog)
* add a test (for the other resolvers they are included, just search for `bcp47` in `translate.determineLocale.spec.js`)
* please rename the resolver name: `lang_only` => `iso639-1`
* the new option should be mentioned in the list of available (incl. an example, i.e. same as in test?): $translateProvider.uniformLanguageTag

knalli:
Superseeded by 1e037ec

PR #1187, opened 2015-09-01
Fix grammar bug

andresgottlieb:


knalli:
@andresgottlieb  Thank you, landed as 664411f

PR #1191, opened 2015-09-02
fix(service): update storage before triggering $translateChangeSuccess

runemoennike:
Triggering $translateChangeSuccess before updating the stored language key causes fallback translation to fail the first time $translate.use() is called. The second time it is called with the same language code, it works. This is because promiseToWaitFor() reads what it thinks is the current language from storage, and then proceeds to ignore that as a fallback language. This is not correct behaviour unless the stored language has been updated first.

This bug occurs if you have languages AA and BB with storage enabled, and AA set up to as the fallback language. Changing to BB from AA causes AA to be ignored in the fallback chain, and all translations fail.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1113,11 +1113,14 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
        */
       var useLanguage = function (key) {
         $uses = key;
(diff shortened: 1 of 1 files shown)


knalli:
Thanks and big thanks for the additional explanation in the commit itself.

Landed as 77dd5a2

PR #1192, opened 2015-09-02
Fix example missing quotes

andresgottlieb:


knalli:
Thanks, landed as 7ddd16b

Issue #1195, opened 2015-09-02
$translate.use() sometimes returns undefined in runTranslate()


fotofinder:
I'm trying to get the value of $translate.use() as soon as possible in my application to set a value on rootScope for later use.
I am using localStorage.
So in my .run() I do: 
```$rootScope.interfaceLang = $translate.use()```

Now I see that this is sometimes ( after reloading the page a couple of times in the browser ) ``ùndefined``` .

In  runTranslate after the 
```javascript
$translate.use(storage.get(key))['catch'](fallbackFromIncorrectStorageValue);
```

line I added this logging statement:

```javascript
console.log("translate has a storage and got:", storage.get(key), " now using ", $translate.use());
```
this is the output:
``translate has storage and got: en  now using  undefined``

If I wrap the logging statement in a fairly long $timeout ( about 1000 millis ) the output is as expected:
``translate has storage and got: en  now using  en``

Unfortunately I have no clue how to mock this in a test.
Any thoughts are welcome.

knalli:
You have to wait for using use() because that is the actual used language. "As soon as possivle" in/after run() leads into undefined.

fotofinder:
Would it be correct to translate a dummy string and to assume that when this translation has completed use() returns the correct value?
```javascript
$translate('DUMMY').then(function(data){
   $rootScope.interfaceLang = $translation.use();
});
```

Or is there an event to listen to?

tspaeth:
Events? Please see http://angular-translate.github.io/docs/#/guide/18_events and check if one of the provided matches your needs.

fotofinder:
see my updated comment

PR #1201, opened 2015-09-04
Emit ready event

fotofinder:
emits an $translateReadyToUse after the promises in $translate.run() are resolved.
This shall indicate that it is now save and reliable to determinate that actually used language with $translate.use().
related to #1195

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1698,14 +1698,17 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
        * When trying to 'use' a language which isn't available it tries to load it
        * asynchronously with registered loaders.
        *
-       * Returns promise object with loaded language file data
+       * Returns promise object with loaded language file data or string of the currently used language.
+       *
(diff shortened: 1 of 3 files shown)


knalli:
Thank you. Please squash the commits together and re-push them again in this PR. Thanks!

Doe the existing events not cover everythings?

tspaeth:
I'm also wondering, why $translateChangeSuccess will not cover it or what the difference it. If there is a "bug" in that, we should preferably discuss or fix it before introducing a 98% clone :-)

fotofinder:
I experienced that $translateChangeSuccess is not fired on application startup.
If it should, I can investigate. Or we fire $translateChangeSuccess also on application start

fotofinder:
@tspaeth : Your are right $translateChangeSuccess is fired on startup. May be I wrongly listened to $translateChangeStart or missed it somehow ( it was late yesterday  ).
Eitherway:
This patch doesn't add any value. I therefore close this PR

PR #1080, opened 2015-06-14
refactor(service): Remove use of legacy methods on $http promises.

realityking:
Also saves a couple of lines of code.

PR #1208, opened 2015-09-12
Handling of the 'useLegacyPromiseExtensions' flag

ghostd:
Angular 1.4.x provides a flag to disable the "old" success/error callbacks of $http. This commit updates the code to use the "then" promise.

### DIFF
--- a/src/service/loader-partial.js
+++ b/src/service/loader-partial.js
@@ -59,10 +59,10 @@ function $translatePartialLoader() {
       $http(angular.extend({
         method : 'GET',
         url: this.parseUrl(urlTemplate, lang)
-      }, $httpOptions)).success(function(data){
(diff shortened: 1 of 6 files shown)


knalli:
@ghostd  Thank you.

However, we have this already included with #1080 with the same reasons.

Perhaps you don't have looked into the `canary` branch (as you should when doing contributions). It will be part of the next minor release.

knalli:
Well.. it could be worth to apply this additional changes right onto the `canary`: https://github.com/angular-translate/angular-translate/pull/1208/files#diff-110b5641457e69e8a6bb537fc6f382beL460

Perhaps you want to make another PR just with the patches for the tests?

ghostd:
Sorry for the noise.

I'm preparing a PR for the tests.

PR #1209, opened 2015-09-13
Using the 'useLegacyPromiseExtensions' flag into the tests

ghostd:
Angular 1.4.x provides a flag to disable the "old" success/error callbacks of $http. This commit updates the test code to avoid regressions.


knalli:
Thanks, landed as 233a012

PR #1220, opened 2015-09-18
feat(service): Add method to provider to get locale at config time.

cthrax:
I am using the angular-google-maps library along with angular-translate, which requires that a config time option be sent to google for which language the maps should be localized too. Rather than re-implement what angular-translate already does, I have added a method on the provider for returning the language that angular-translate is at least going to try first.

Of course this method will only provide a proposed language, because it's not until the languages are fetched that angular-translate can fully determine what language will be used, but I think that fallback mechanism can be handled differently, due to a different set of translations, for the library consuming this value.

It seems like it breaks the chaining that is presently in all the provider methods, but I didn't see a way around that without exposing some of the local variables as properties. Thoughts?

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -845,6 +845,25 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
 
   /**
    * @ngdoc function
+   * @name pascalprecht.translate.$translateProvider#getProposedLanguage
+   * @methodOf pascalprecht.translate.$translateProvider
+   *
+   * @description
+   * Retrieves the language will initially be fetched, this could have been set through the use method or determined
+   * by {@link pascalprecht.translate.$translateProvider#determinePreferredLanguage determinePreferredLanguage}.
+   *
+   * This language may not be available and another fallback will actually be used, but that cannot be determined until
+   * runtime. The intent of this method is to give a hint at config time as to what language should be used, mostly so
+   * other libraries that are translated outside of angular-translate can stay in lock step with angular-translate's language.
+   *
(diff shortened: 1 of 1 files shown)


knalli:
@cthrax  Thank you.

@angular-translate/core  What do you think? Should it be perhaps flipped?

PascalPrecht:
@knalli what do you mean by "flipped"?

knalli:
`$preferredLanguage` before `$uses`?

But I think I was wrong. $uses should be a higher priority.

cthrax:
So is there any movement on this? Do you need something else from me?

tspaeth:
I'm asking myself if there really is any business value to the module as during config phase $translate.use and $translate.preferredLanguage both return values if queried - or am I just scanning our sourcecode quickly wrong in github.com? :-)

cthrax:
@tspaeth $translate is not available during the config stage, only $translateProvider. The "use" method of $translateProvider returns the private variable $uses, which is set by "use". If one is detecting the language, then $uses is not set until $translate is instantiated, so the "preferredLanguage" is calculated, but the $translateProvider does not expose that preferred language, likely due to the fact that it may not be the language used based on available translations.

So in short, no it does not return values at config time.

tspaeth:
$translateProvider does expose preferredLanauge - or not?

See: https://github.com/cthrax/angular-translate/blob/feature/configTimeLocale/src/service/translate.js#L396

cthrax:
@tspaeth ah, I am/was using an old version I guess. This appears to have fixed what I wanted. https://github.com/angular-translate/angular-translate/commit/6014a813c6495b2a2d86215acdc2f4b8d690487b

Thanks for pointing it out and I retract my PR.

PR #1225, opened 2015-09-23
fix(service): should not abort fallback languages

yannickglt:
The loadAsync method is not aborted anymore to avoid issues with loading of fallback languages. To fix the synchronization problem solved with #1070, we just check if a language is the last requested one to use it.


acourtiol:
:+1:

knalli:
landed as cc410b1

PR #1232, opened 2015-09-26
fix subheadline

piotrowski:


knalli:
Thanks, landed as d34f5c7

PR #1234, opened 2015-09-26
fix 'useful resources'

piotrowski:
- remove anchor to specific comment on technpol.wordpress.com link
- remove neoskop article as it still uses `$translate.uses()`


knalli:
Thanks, applied for all languages and landed as 7c5fff1

PR #1237, opened 2015-10-01
feat(directive): Add translation IDs to the scope

skovalyov:
This change allows to use translation IDs extracted by `translate`
directive in other directives.

For example, if you are developing a translation editor, you can create
a directive, let’s call it `angular-translate-edit`, with `restrict:
‘A’`, that would get translation IDs defined for the angular-translate.

This gives you the possibility to avoid some complex code duplication
with all different options to define translation IDs, including
interpolation, what is already done in the `translate` directive.

As a part of this change one duplicated test is removed that uses
`$rootScope.translationIds` due to the name conflict in the
prototypically inherited directive scope. The same test already exists
and it just uses `$rootScope.translationId` instead.

### DIFF
--- a/src/directive/translate.js
+++ b/src/directive/translate.js
@@ -126,7 +126,7 @@ function translateDirective($translate, $q, $interpolate, $compile, $parse, $roo
         scope.preText = '';
         scope.postText = '';
         scope.translateNamespace = getTranslateNamespace(scope);
-        var translationIds = {};
+        scope.translationIds = {};
 
         var initInterpolationParams = function (interpolateParams, iAttr, tAttr) {
           // initial setup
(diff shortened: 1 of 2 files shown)


tspaeth:
Hm, isn't putting all translationIds themselves into scope putting  a lot of pollution to the scope variables/listeners? Or am I just interpreting it a bit wrong?

skovalyov:
@tspaeth You are right. But there are `interpolateParams`, `preText`, `postText` and `translateNamespace` already there from this directive. And I am not insisting that my solution is right, but what would be the other way to get the specific element `translationIds` from another directive? The reason I need it is mentioned in the commit comment - I am creating a sort of inline translation tool for the project where we use `angular-translate` and I would like to share IDs without ID duplication or duplicating the same code for extracting those from the `angular-translate` which would be also not reliable when `angular-translate` evolves further.

skovalyov:
My inline translation tool directive now starts like this:

```javascript
return {
  restrict: 'A',
  link: function(scope, element, attrs) {
      var onClick = function(event) {
        event.stopImmediatePropagation();
        var translationId = attrs.translate;
        ...
        ...
        ...
```

So basically I'm getting translation ID from translate attribute only. But it is only one of the option provided by `angular-translate`. And I would like to support all options. But then I need to duplicate a lot of `angular-translate` code including translation IDs interpolation, both in the element text and attributes. It is of course possible, but does not seem to be a reliable solution. That's why I consider there should be some nice way to share translation ID from `angular-translate` translate directive.

knalli:
Sorry for the delay. I have a problem adding so much data in scope (which will be picked up by the watcher/scope-chain of AJS) only for a may-have or optional feature. Mostly will not use this.

Unless we can make this being opt-in, I don't see this will be good change :/

skovalyov:
Ok, I see. Probably I need to find an alternative way to achieve the goal behind this proposed change.

PR #1259, opened 2015-10-13
Meteor publish package for atmospherejs.com

silintzir:


Heinerion:
+1

package.js
```diff
@@ -0,0 +1,22 @@
...
+	version: version,
+	summary: summary,
+	git: gitLink,
+	documentation: documentationFile
+});
+
+Package.onUse(function(api){
+	api.versionsFrom(['METEOR@0.9.0', 'METEOR@1.0']); // Meteor versions
+	api.use('angular:angular@1.2.26', where); // Dependencies
+	api.addFiles('src/translate.js', where); // files in use
```
knalli: That does not look like a good entrypoint for the package. In general, we publish only the built artifacts, but no source files.

knalli:
In which way will be the package announced (both published and republished again) in the target repository?

Heinerion:
The package would be published as = `silintzir:angular-translate`, where `silintzir` is the name of @silintzirs Meteor Account.
If you'd like to see your name, you'd need to register to meteor.
After that you could release the package as `knalli:angular-translate` or `angular-translate:angular-translate`. Whatever name you chose for your account.

knalli:
That is not what I want to know. Which hooks will be available? How do resolve your local dependency? Where are the generated sources? What about git tagging?

Heinerion:
I'm sorry, I got the question wrong.
I have to admit I didn't pay really much attention to @silintzir s package.js, I would just love to include the real and up-to-date angular-translate package in meteor.

The idea would be to ship the *generated* js file unminified as package.
Not the sources, but the already built script.

In a meteor project, this file would be used unminified while debugging and would be minified with the rest of the application in production mode.

If I misunderstood your questions again, please refer to [Angular-Meteor](http://angular-meteor.com/tutorials/angular1/3rd-party-libraries) and find the headline **Solution 1**

knalli:
If I understand this correctly, we have to publish each version and each package manually for meteor. Hm.. I don't like another package system besides bower (currently IMHO should) and npm (prefer).

Also I don't understand why you are not using npm? Simply `npm init` and `npm install angular-translate -D`.

Additional resources I've found so far:
* meteor-bower https://github.com/mquandalle/meteor-bower
* meteorhacks npm https://github.com/meteorhacks/npm
* meteor-npm https://www.npmjs.com/package/meteor-npm

knalli:
I'll close this.

I want to keep the amount of spec files to manage as small as possible. At the moment, we have the following ones:

* `package.json` for NPM -- This is mandatory and the de facto standard for JavaScript modules
* `bower.json` for Bower -- Although NPM beats  bower in several points, a lot of people are still using Bower. And well, it was and actually is somehow a broad-wide 2nd de facto standard.
* `component.json` Well.. it was a great module loader, but it is deprecated now. It is still there because it does not need to be maintained.

The last point is important: these things have to be maintained, to be managed. The spec file `package.js` above is a proprietary one and generated for single vendor/platform. And it has to be managed on each release. 
It even includes details about meteor (valid for 0.9, 1.0). We don't know this, we will not care of this (probably).

Fortunately, there are already bridges for Meteor developers in the existing ecoystems (npm, bower) of modules. Unless there is a serious point why you cannot use neither meteor-bower no meteor-npm, we don't have to add this here.

Heinerion:
Alright, thanks for your effort anyway!

PR #1265, opened 2015-10-16
feat(service): Add `negotiateLocale` method

moshen:
Add the `$translate#negotiateLocale` method.

Exposing `negotiateLocale` enables testing language keys without calling `$translate#use`.  This is particularly helpful in cases where languages and aliases are known and registered on load.

We have several sites which route based on language key, being able to test validity directly would be beneficial.


knalli:
@moshen  Thank you, landed as 9247000

moshen:
:beers: Thanks!

PR #1299, opened 2015-11-17
Update package.json to add dependency to angular.

wlepinski:
This is mainly because some package managers introspect this property to also get all dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,9 @@
     }
   ],
   "license": "MIT",
+  "dependencies": {
+    "angular": "^1.4.7"
+  },
   "devDependencies": {
     "bower": "^1.3.12",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -34,6 +34,9 @@
     }
   ],
   "license": "MIT",
+  "dependencies": {
+    "angular": "^1.4.7"
```
tspaeth: Why should it be ^1.4.7? Doesn't that exclude 1.2.x and 1.3.x as dependencies?

knalli: Yep. It would. It must be the same range like in the other ones.

knalli:
I've had to changed this, landed as 475a9b6

wlepinski:
:+1:

PR #1308, opened 2015-11-20
Fix for changing `translation-default` attr value

bildja:
If we have something like this:
```
<span translate="{{not_existing_translation_key}}" translate-default="{{something.isChanging}}"></span>
```

And if the value of `something.isChanging` changes over time, the value does not update accordingly


knalli:
@bildja  Yes, looks good. Landed as ea94acd

PR #1312, opened 2015-11-25
chore(): add support of default translation in missingTranslationHandler

LoicMahieu:
We can both define a default translation and a missing translation handler. In my case, I need to treat the default translation in the missing handler. This PR add the `defaultTranslationText` parameter to the handler.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1342,14 +1342,16 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
        * Translate by missing translation handler.
        *
        * @param translationId
+       * @param interpolateParams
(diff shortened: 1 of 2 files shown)


tspaeth:
Just a general question: Do you have several DefaultTranslations? And are they language dependent? And what do you do further with them after being in the result of the missingHandler?

LoicMahieu:
I'm late!
My purpose is only to allow us to use both a missing handler and the default text.
I needed to add logic on default translation and return a modified version.
This PR gives more scope to the `missingTranslationHandler`.

I hope I answer your question. Thanks

knalli:
@LoicMahieu thank you, I've picked this as 8c5044c.

Issue #1101, opened 2015-06-24
Sanitization issue with arabic characters


badr-ghatasheh:
Hi,

There is an issue with applying 'sanitize' strategy while using Arabic characters

for example sanitizing the word 'بدر' will print out the following on the page:
&#1576;&#1583;&#1585;

It turns out the reason this happens is that you use element.text() to fill the element with the sanitized value in the following function (/service/sanitization.js):

var htmlEscapeValue = function (value) {
    var element = angular.element('<div></div>');
    element.text(value); // not chainable, see #1044
    return element.html();
  };

replacing element.text(value) by element.html(value) worked for me.

Can this be fixed?
Thank you

n00dl3:
same problem with any utf8 character, "éèë..."

tspaeth:
This is somehow a thing we should discuss. But I'm just myself wondering, why "simple UTF*" conversion fails in the jqLite-elements this way. Needs a bit of investigation I guess.
But as it impacts also the security a bit on sanitizing elements, my first suggestion would be to introduce a "trusthtml" sanitize strategy as an additional option for those really wanting to open that door :-). 

Somehow a parallel issue exists also on angular itself to watch: https://github.com/angular/angular.js/issues/4790

@knalli what do you think?

knalli:
I'm a little bit confused.

@badr-ghatasheh: You are writing you are faced with this issue when

> applying 'sanitize' strategy

but you are actually not meaning `sanitize` but `escape, do you? Because `htmlEscapeValue` will not be used when `sanitize` is activated (it is not required).

Remarking your "fix": That's not a fix, that would make the whole thing ad absurdum: http://plnkr.co/edit/cDCBXlWVgYtny9ciJGV9?p=preview It only works using standard AJS 1.3+ with a $sce running in the background (i.e. `ng-bind-html`).

@tspaeth I've listed the different kinds of: http://plnkr.co/edit/cDCBXlWVgYtny9ciJGV9?p=preview I'm quite not sure what this does mean for us right now.

badr-ghatasheh:
Hi @knalli 

I Updated the issue, I seem to have mixed the function names.

However, the issue still applies to 'sanitize' strategy and utf8 characters.

samal84:
More people having this issue (including us): http://stackoverflow.com/questions/31002499/angular-translate-sanitisation-fails-with-utf-characters

knalli:
[...]

PR #1313, opened 2015-11-30
Add desc and fix of an error related to this doc

hobbysoldat:
There is a bug in angularjs or angular translate or some misscommunication between the two frameworks #1101 where some sanitize modes do not work correctly. Instead of letting our users run into the knife and hurt until they search for the bug in the issues list, I added an intermediate solution from the issue.


knalli:
@hobbysoldat  Thanks.

On hold, we should add the notes for other languages (as much as possible) as well.

hobbysoldat:
OK. I added a german version too!

tspaeth:
"There is a bug in angularjs #1101" - does this give the correct information? #1101 is the angular-translate issue. But is there another one directly in AngularJS?

hobbysoldat:
Well according to the description in the bug report the bug is in the behaviour of angular $sanitize. Maybe angular translate is using it wrong.

knalli:
Thank you, landed as c66e5f4

PR #1316, opened 2015-12-01
Apply notFoundIndicators only when all configured language checked and translation isn't found in $translate.instant method

### LINKED ISSUE #1314: $translate.instant doesn't apply fallback language translation in case of translateNotFoundIndicators usage since 2.6.0
Translation table contains two languages with codes 'en' and 'de', using 'en' as fallback language and 'de' as current.Translation for key exists only for fallback language. Also translateNotFoundIndicators are configured. But when I attempt to get translation using translate filter, I get translation key, which is wrapped into indicators. Expectation is to get fallback language translation.
Issue reproduced in the plunker:  [example](http://plnkr.co/edit/zndZQjPg8AWfnEDviiLY?p=preview).

shkrobova:
fix($translate): apply notFoundIndicators only when all configured language checked in $translate.instant method

Fixes #1314

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1969,8 +1969,6 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
           if ($translationTable[possibleLangKey]) {
             if (typeof $translationTable[possibleLangKey][translationId] !== 'undefined') {
               result = determineTranslationInstant(translationId, interpolateParams, interpolationId);
-            } else if ($notFoundIndicatorLeft || $notFoundIndicatorRight) {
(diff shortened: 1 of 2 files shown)


tspaeth:
Hi! Thanks for the fast contribution. 
functionally looks good to me and formally nearly perfect :-)

shkrobova:
Hi! Thanks for the fast commit check. Can I expect that it will be included in the next release?

Behzadkhosravifar:
Hi, thanks for your guidance.

Behzad Khosravifar

knalli:
Thank you very much.

landed as 25b13c4

PR #1328, opened 2015-12-10
feat(service): force language used for translating

odedniv:
support multiple languages in the same template (discussed in #500) by
allowing to force the used language through:

* a parameter in $translate() and $translate.instant().
* scope variable for the translate directive, including a directive to
  change that variable.
* a parameter in translate filter.

### DIFF
--- a/src/directive/translate.js
+++ b/src/directive/translate.js
@@ -5,7 +5,7 @@ angular.module('pascalprecht.translate')
  * @requires $compile
  * @requires $filter
  * @requires $interpolate
- * @restrict A
+ * @restrict AE
  *
  * @description
  * Translates given translation id either through attribute or DOM content.
@@ -259,7 +259,7 @@ function translateDirective($translate, $q, $interpolate, $compile, $parse, $roo
               translationId = translateNamespace + translationId;
             }
 
-            $translate(translationId, interpolateParams, translateInterpolation, defaultTranslationText)
+            $translate(translationId, interpolateParams, translateInterpolation, defaultTranslationText, scope.translateLanguage)
               .then(function (translation) {
                 applyTranslation(translation, scope, true, translateAttr);
               }, function (translationId) {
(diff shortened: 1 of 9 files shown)


knalli:
@odedniv Great addition.  Could you please add at least two tests (both directive and filter) with an overridden language?

src/directive/translate.js
```diff
@@ -5,7 +5,7 @@ angular.module('pascalprecht.translate')
  * @requires $compile
  * @requires $filter
  * @requires $interpolate
- * @restrict A
```
knalli: why?

odedniv: Well actually I just saw that it was wrong (as I created a new directive), so I fixed it... Any reason not to?

I know from my experience that it allows elements, and you can see that from the [code](https://github.com/angular-translate/angular-translate/blob/master/src/directive/translate.js#L107).

knalli: Opps.. didn't read carefuly enough. Ok!

knalli:
Also please rebase. Looks there are conflicts..

odedniv:
Yes, just saw it, already done. It seems to have failed a test, I'm working on it.

I also looked at the tests while implementing and tried to figure out where to squeeze mine, but failed to... I would love it if I could get some guidance there (or a pull request to my branch or something).

Tests are always the most cumbersome part of a repository :smile:

knalli:
For tests: like in https://github.com/angular-translate/angular-translate/blob/ea94acd0bb16c8ffb286ec5676d7f6d35933e5e6/test/unit/directive/translate.spec.js#L594-L661 -- make such a new "scope" and put in your tests. Same for the filter..

odedniv:
Should I insert/fix the tests with commit --amend or new commit?

knalli:
--amend sounds good :+1:

odedniv:
Just fixed failing tests, new tests will come tomorrow (will update with a comment).

odedniv:
Added tests. Not sure I'm using language negotiation properly there, please go over it before pulling.

knalli:
@odedniv  Excellent work! :shipit:

knalli:
Landed as e591462 -- thank you!

PR #1333, opened 2015-12-14
fix(service): partial loader service refetches list of parts

### LINKED ISSUE #1326: When using partial translations and multiple translate.refresh() are called not all translations are loaded
When I call $translatePartialLoader.addPart('...') and translate.refresh() in a short space of time the full list of translations is not always loaded in, this leads to the end user being shown missing translations.

The issue seems to happen when you add a part then refresh translations then add another part and refresh it before the previous refresh has completed. If the second refresh is completed then the first is completed then the results of the second refresh are ignored as the first one (with less data) overwrites the data from the second (which has the extra data).

Hopefully this example makes it clearer.
http://codepen.io/antisback/pen/xZGKQK?editors=101
You can toggle the translation working by commenting out the timeout near the end to simulate the first refresh resolving after the second.

cleavera:
Ensures that the partial loader service refetches the list of parts when its promise resolves in case the list of parts have changed in the intervening time

Fixes #1326

### DIFF
--- a/src/service/loader-partial.js
+++ b/src/service/loader-partial.js
@@ -308,6 +308,7 @@ function $translatePartialLoader() {
       $q.all(loaders)
         .then(function() {
           var table = {};
+          prioritizedParts = getPrioritizedParts();
           angular.forEach(prioritizedParts, function(part) {
             deepExtend(table, part.tables[options.key]);
           });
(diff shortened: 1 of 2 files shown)


knalli:
@cleavera  Thank you, landed as 069eafd

PR #1339, opened 2015-12-21
fix (translate directive): do not translate content if empty

### LINKED ISSUE #1338: translate attributes only for element with whitespace "content"
I created a little [jsfiddle](https://jsfiddle.net/szqqLq6q/3/) that shows the issue. Angular translate removes the icon from the link as it expects that it should translate the content of the link. But that's not the case, it should only translate the attribute. 

The issue disappears if the content is not formatted, e.g. when we define the link like this:

```html
<a href="#" translate translate-attr-title="Home"><i class="fa fa-home"></i></a>
```

the point is, that the text of an element with nicely formatted child elements is not undefined but an empty string. 

Solution. Do no test if the translation id is undefined but test if the translation id is falsy. 
I expect that we nee to adjust the condition on line 245 of the translate directive.

If you agree with the provided solution, then I will prepare a pull request that adjust the behaviour.

DatenMetzgerX:
Do not translate the content of an element if the element has translate-attr-* attributes and
the element-content is empty. E.g. do not translate the content of 

```js
<a href="#" translate translate-attr-title="Home">
     <i class="fa fa-home"></i>
</a>
```

fixes #1338

### DIFF
--- a/src/directive/translate.js
+++ b/src/directive/translate.js
@@ -173,7 +173,8 @@ function translateDirective($translate, $q, $interpolate, $compile, $parse, $roo
                 });
               }
             } else {
-              translationIds.translate = iElementText;
+              // do not assigne the translation id if it is empty.
+              translationIds.translate = !iElementText ? undefined : iElementText;
             }
           } else {
             translationIds.translate = translationId;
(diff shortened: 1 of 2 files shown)


knalli:
@DatenMetzgerX  And thank you again. Well looking contribution, including tests. Perfect.

Landed as cafc0cf

PR #1341, opened 2015-12-25
chore(loaders): reuse deferred objects instead of creating new ones

michaelbazos:
The purpose of this PR is to remove the promise antipatterns from the loaders.


knalli:
@michaelbazos  Thank you, perfect. Landed as-is as 3344cff

PR #1343, opened 2015-12-28
fix grunt default task didn't match contribution documentation

tahoemph:
Default task did both a lint and a full test run instead of just a
lint as specified by the documentation.  Change that to only do a
lint.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -652,7 +652,7 @@ module.exports = function (grunt) {
   });
 
 
-  grunt.registerTask('default', ['jshint:all', 'karma']);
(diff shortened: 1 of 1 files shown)


knalli:
@tahoemph  Thank you for pointing out that issue. However, I'm updating the documentation being correct. The code was right already.

tahoemph:
I originally made that change but thought it odd that the default grunt task did more test then 'test' task itself.  One thing you might consider is changing the default task to do jshint:all and karma:unit as this fits fairly close with the description of the default action in the documentation.

Issue #1359, opened 2016-01-12
Translate promise should be rejected when translationId is not found and missingTranslationHandlerTranslation is defined


pdore-netfore:
https://github.com/angular-translate/angular-translate/blob/master/src/service/translate.js#L1487

knalli:
Discussion => #1361

PR #1361, opened 2016-01-13
Ensure translate promise is rejected when translation is missing.

pdore-netfore:
Reject translate promise when translation is missing and a custom 'missingTranslationHandlerTranslation' is provided.

Solves https://github.com/angular-translate/angular-translate/issues/1359

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1484,7 +1484,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
             if (defaultTranslationText) {
               deferred.resolve(defaultTranslationText);
               } else {
-                deferred.resolve(missingTranslationHandlerTranslation);
+                deferred.reject(missingTranslationHandlerTranslation);
               }
           } else {
             if (defaultTranslationText) {


knalli:
@pdore-netfore  Thank you for contribution.

@angular-translate/core  It sounds good, but I'm wondering..

* no test is failing? looks like a lack of coverage.. :(
* is there any reason why we should NOT accept this change?

knalli:
Ah, just realized the thing again. `$missingTranslationHandlerFactory` is an alternate way of resolving the translation. `resolve` is correct, `reject` not. You don't have to add such a handler (via `$translateProvider.useMissingTranslationHandler()`).

@pdore-netfore  What is your actual indent?

tspaeth:
@knalli I was also asking myself these questions you had :-)

pdore-netfore:
When supplying your own translation handler, the promise should be rejected when a key is missing (the same way as it does when using the default handler)

pdore-netfore:
Please merge if reviewed

knalli:
We will not merge this.

As I said already: missingTranslationHandlerFactory is an alternative way of providing translation texts. The result of this is a valid translation which means the result is resolving one. Otherwise the other concepts like fallback languages will not work as expected.

> (the same way as it does when using the default handler)

The default handler rejects in the situation of a missing translation. The main purpose of the missingTranslationHandlerFactory is to fix this.

Perhaps we need something else, but the existing behavior works as expected under the context of the other features.

PR #1377, opened 2016-01-25
Allow 1.4 and 1.5 beta releases to satisfy angular dependency

df-dylan:
Without this, npm won't allow a beta angular version in a parent project to satisfy angular-translate's angular dependency. See https://docs.npmjs.com/misc/semver#prerelease-tags

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,6 @@
     "plato": "^1.5.0"
   },
   "dependencies": {
-    "angular": ">=1.2.26 <=1.5"
+    "angular": "^1.2.26 || >= 1.4.0-beta.0 || >= 1.5.0-beta.0"
   }
 }


jcwatson11:
+1 We need this. Please review and merge! Thanks!

knalli:
I don't get the problem.

Given the sample package.json

```json
{
  "name": "q",
  "version": "1.0.0",
  "devDependencies": {
    "angular": "1.5.0-beta.0",
    "angular-translate": "^2.9.0"
  }
}
```

I'm getting this dependency stack

```
├── angular@1.5.0-beta.0
└─┬ angular-translate@2.9.0
  └── angular@1.5.0
```

Yes, there is an additional dedicated angular inside of angular-translate. However, since we don't use npm internally (except for tooling at the developing stage), this is an acceptable tradeoff.

Did I missed something else?

Additionally, 1.5 has been released yesterday. And the 1.5 release is already in the range.

knalli:
Resolved in latest 2.9.0 (fixup 2.9.2)

jcwatson11:
I agree with knalli. I thought this was a problem at first too, but later deleted my comment from this thread because I understood what knalli just explained.

PR #1380, opened 2016-01-27
Add support for promise result in storage.get

vtrifonov:
Adding possibility to use a storage provider returning promise from it's get function, so that it can make an asynchronous call to an endpoint to get the language.


knalli:
Perhaps it is better to use $q.resolve/$q.reject automatically checking for promise values.

knalli:
@vtrifonov  Thank you for making the pull request and the support for promise based caches. Please follow our contribution guidelines. Therefore we cannot accept the PR.

However, we would welcome a re-pull!

PR #1387, opened 2016-02-01
fix(directive): reduced number of watchers for translate-language directive

cstefanache:
- replaced 'translateLanguage' listener from translate.js with eventListener in order to reduce number of watchers on scope.
- introduced listener only when the translate-language directive is used
- updated unit tests with following scenarios:
 1. translate-language directive used on isolated scope should not change the parent translate language
 2. allow changing the translations for translate-language isolated DOM at runtime
 3. language change triggered on the application should not affect translate-language isolated DOM translations
- introduced demo example for the translateLanguage (ex14_translate_language.htm). Initial number of watchers on demo app before fix: 14, after fix: 12. Translate.js initially added 4 extra watchers for all isolated scopes. The demo application has 2 dom elements that are using translate-language directives thus, after the change, 2 extra watchers (instead of 4) are registered on the scope.

### DIFF
--- a/src/directive/translate-language.js
+++ b/src/directive/translate-language.js
@@ -54,8 +54,13 @@ function translateLanguageDirective() {
     scope: true,
     compile: function () {
       return function linkFn(scope, iElement, iAttrs) {
+
         iAttrs.$observe('translateLanguage', function (newTranslateLanguage) {
           scope.translateLanguage = newTranslateLanguage;
+        });
+
+        scope.$watch('translateLanguage', function(){
+          scope.$broadcast('translateLanguageChanged');
         });
       };
     }
@@ -63,4 +68,3 @@ function translateLanguageDirective() {
 }
 
 translateLanguageDirective.displayName = 'translateLanguageDirective';
-
--- a/src/directive/translate.js
+++ b/src/directive/translate.js
@@ -245,7 +245,6 @@ function translateDirective($translate, $q, $interpolate, $compile, $parse, $roo
         // Master update function
         var updateTranslations = function () {
           for (var key in translationIds) {
-
             if (translationIds.hasOwnProperty(key) && translationIds[key] !== undefined) {
               updateTranslation(key, translationIds[key], scope, scope.interpolateParams, scope.defaultText, scope.translateNamespace);
             }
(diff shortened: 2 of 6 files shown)


src/directive/translate-language.js
```diff
@@ -54,13 +54,17 @@ function translateLanguageDirective() {
     scope: true,
     compile: function () {
       return function linkFn(scope, iElement, iAttrs) {
+
         iAttrs.$observe('translateLanguage', function (newTranslateLanguage) {
           scope.translateLanguage = newTranslateLanguage;
         });
+
+        scope.$watch('translateLanguage', function(){
+          scope.$broadcast('translateLanguageChanged');
```
knalli: Potentially this is a lot of work.. broadcast will notify every child scope.

src/directive/translate.js
```diff
@@ -303,7 +302,9 @@ function translateDirective($translate, $q, $interpolate, $compile, $parse, $roo
         if (translateValuesExist || translateValueExist || iAttr.translateDefault) {
           scope.$watch('interpolateParams', updateTranslations, true);
         }
-        scope.$watch('translateLanguage', updateTranslations);
+
+        // Replaced watcher on translateLanguage with event listener
+        var unbindTranslateLanguage = scope.$on('translateLanguageChanged', updateTranslations);
```
knalli: So, in the end this all here: you have replaced this particular watcher (old) and replace it with a new event listener. I'm still looking for the 2 watcher reduction..

cstefanache: Regarding the watcher reduction: with each isolated scope, the previous version would register one 'translateLanguage' watcher regardless the presence of the translate-language directive usage. The commit contains a new demo file: ex14_translate_language.htm that uses twice the translate-language directive (previous implementation: 14 watchers, current implementation: 12 watchers) by avoiding adding 'translateLanguage' watchers for the following DOM node elements:
<h2 translate="SUBHEADER">Subheader</h2>
<p translate>HTML_KEYS</p>

knalli: Ah.. because of the additional `translate-language` in in the h4. I didn't got this. Thanks.

src/directive/translate-language.js
cstefanache: The purpose of notifying child scopes is to make sure that all child nodes will be updated accordingly. I have added two unit tests to the existing stack that are testing the following:
** it('should not change parent scope language', ...) **- the current translateLanguage change should not affect translations in the parent scopes
** it('should not be changed by parent scope language'** - preserve language preference below the current scope

knalli:
@cstefanache  Thank you for this great addition, picked as 961fc92.

Issue #1396, opened 2016-02-10
Infinite loop when translate params is a recursive object


dceddia:
Passing a self-referential object as the translate params causes an infinite loop when using the `sanitizeParameters` sanitization strategy:

```
RangeError: Maximum call stack size exceeded
    at String.replace (native)
    at encodeEntities (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.0/angular-sanitize.js:458:5)
    at Object.chars (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.0/angular-sanitize.js:520:15)
    at htmlParser (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.0/angular-sanitize.js:353:36)
    at http://ajax.googleapis.com/ajax/libs/angularjs/1.4.0/angular-sanitize.js:155:7
    at htmlSanitizeValue (http://cdnjs.cloudflare.com/ajax/libs/angular-translate/2.9.0/angular-translate.js:304:12)
    at mapInterpolationParameters (http://cdnjs.cloudflare.com/ajax/libs/angular-translate/2.9.0/angular-translate.js:319:14)
    at http://cdnjs.cloudflare.com/ajax/libs/angular-translate/2.9.0/angular-translate.js:312:31
    at Object.forEach (http://ajax.googleapis.com/ajax/libs/angularjs/1.4.0/angular.js:350:20)
    at mapInterpolationParameters (http://cdnjs.cloudflare.com/ajax/libs/angular-translate/2.9.0/angular-translate.js:311:15)
```

The issue seems to be that `mapInterpolationParameters` calls itself without checking for cycles.

See the JSBin here: http://jsbin.com/cocosiwura/1/edit?html,js,output (check the console)

knalli:
If you shoot yourself in the foot, it will probably hurt. :)

Anyway: A call stack error is not good. I've added some lines holding the state (the stack), and print out a proper message: See here in action: http://jsbin.com/vogatohote/1/edit?html,js,output

dceddia:
Haha, well, yes :)

Would it not be better to simply ignore the recursive references? As a user I imagined that it would only sanitize the params that my translation string required (which wasn't the case, but still). I think the "principle of least surprise" would be to just sanitize everything while removing the recursive references.

See #1401

knalli:
The error has been resolved in 2.9.2 already, but we have an open issue regarding the reaction.

PR #1401, opened 2016-02-14
fix(service): remove recursive references during sanitization

dceddia:
Related to #1396

Replace recursive references with `undefined` rather than throwing an error.

### DIFF
--- a/src/service/sanitization.js
+++ b/src/service/sanitization.js
@@ -247,7 +247,8 @@ function $translateSanitizationProvider () {
         stack = [];
       } else {
         if (stack.indexOf(value) > -1) {
-          throw new Error('pascalprecht.translate.$translateSanitization: Error cannot interpolate parameter due recursive object');
+          // Replace recursive object refs with undefined
+          return;
         }
       }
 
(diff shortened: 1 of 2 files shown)


knalli:
Well, I'm not sure it is a good idea allowing "recursive object references". Putting this into a AJS scope will result into similar issues, too (due AJS' expression watchers). A well known object sample is a HTML element object which references itself in the children.parent references.

Said that, I have the following addtions:

1. Make the `undefined` result a opt-in feature (i.e. `$translateProvider.dontThrowExceptionOnSelfReferencingInterpolationParams(true) .. but shorter! :)).
2. Instead of resulting into an `undefined`, put at least an optional logging (i.e. `$log.warn(...)`).

knalli:
@tspaeth wdyt?

tspaeth:
I'm a bit far away from the topic itself. 
But with a bit reading into the issue, I'll agree that changing the behavior to undefined without a switch is somewhat a breaking change (I can't judge on the side effects just by reading... ). So I would also say: yup, at least as an opt-in feature / toggle.

knalli:
Close this.

The thrown error is not acceptable and must be handled. We only catching the *already existing error* avoiding further escalation which results probably into undistinguishable error stack traces in a complete app.

Issue #1415, opened 2016-02-24
fallbackLanguage broken after upgrade to 2.9


sebek64:
### Subject of the issue
After upgrading from angular 1.4 to 1.5 and angular-translate from 2.8 to 2.9, language switching to "fallbackLanguage" stopped working.

### Your environment
* version of angular-translate: 2.9.2
* version of angular: 1.5.0
* which browser and its version: Firefox, Chrome

### Steps to reproduce

In our application, language switching stopped working in a strange way. Switching among all languages works, but we cannot switch to 'en'. After some investigation, we found that dropping

$translateProvider.fallbackLanguage('en');

fixes the switching problems. Here is a part of our code:

    .config(function ($translateProvider) {
      $translateProvider.useLocalStorage();
      $translateProvider.useSanitizeValueStrategy('sanitizeParameters');
      $translateProvider.useMessageFormatInterpolation();
      $translateProvider.preferredLanguage('en');
      $translateProvider.fallbackLanguage('en');
      $translateProvider.uniformLanguageTag('bcp47');
      $translateProvider.determinePreferredLanguage();
      $translateProvider.registerAvailableLanguageKeys(['en', 'cs', 'pl'], {
        // en
        'en-AU': 'en', 'en-BZ': 'en', 'en-CA': 'en', 'en-CB': 'en', 'en-GB': 'en', 'en-IE': 'en', 'en-JM': 'en',
        'en-NZ': 'en', 'en-PH': 'en', 'en-TT': 'en', 'en-US': 'en', 'en-ZA': 'en', 'en-ZW': 'en',
        // cs
        'cs-CZ': 'cs',
        //pl
        'pl-PL': 'pl'
      });

      $translateProvider.useLoader('$translatePartialLoader', {
        urlTemplate: 'modules/{part}/{lang}.json',
        loadFailureHandler: 'SITranslationErrorHandler'
      });
      $translateProvider.useLoaderCache(true);
    })

    .run(function ($rootScope, $stateParams, $translate, $translatePartialLoader){
      $rootScope.$on('$stateChangeSuccess', function () {
        if ($stateParams.lang !== undefined){
          $rootScope.activeLang = $stateParams.lang;
          $translate.use($stateParams.lang);
          $translatePartialLoader.addPart('translation');
          $translate.refresh();
        }
      });
    });

Another way how to fix switching (but without dropping fallback language) is to move fallback language to .run() block:

$translate.fallbackLanguage('en');

[...]

PR #1419, opened 2016-02-26
fix(service): make the fallback $uses / $translate.use work in a corr…

tspaeth:
…ect manner

test(fallback): add a test case for the scenario described in #1415

@knalli have a short recheck please tests look good, my personal test case locally also (demo application I'm running)

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1758,8 +1758,8 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
 
         // if there isn't a translation table for the language we've requested,
         // we load it asynchronously
+        $nextLang = key;
         if (($forceAsyncReloadEnabled || !$translationTable[key]) && $loaderFactory && !langPromises[key]) {
-          $nextLang = key;
           langPromises[key] = loadAsync(key).then(function (translation) {
             translations(translation.key, translation.table);
             deferred.resolve(translation.key);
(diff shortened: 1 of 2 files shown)


knalli:
Looks good for me.

src/service/translate.js
```diff
@@ -1758,8 +1758,8 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
 
         // if there isn't a translation table for the language we've requested,
         // we load it asynchronously
+        $nextLang = key;
```
knalli: This already fixes some issues I guess.

src/service/translate.js
```diff
@@ -1780,10 +1780,10 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
           // we are already loading this asynchronously
           // resolve our new deferred when the old langPromise is resolved
           langPromises[key].then(function (translation) {
-            if (!$uses) {
-              useLanguage(translation.key);
+            if ($nextLang === key) {
```
knalli: Hm.. `translation` will be ignored completely?

tspaeth: Just took it as it was below - could we have the situation where key is already running out if translation.key? Maybe I was just misreading it.

src/service/translate.js
tspaeth: Yup. Kills some of the fallback-async issues :smile:

src/service/translate.js
knalli: Should not be `$nextLang === translation.key`?

tspaeth: That is what I meant with the last comment: I first thought, that there might not be a difference in between key and translation.key , but due to a very slow reply, this could possibly happen after rethinking... or not?

knalli:
Picked as 7e71a5a w modification

PR #1421, opened 2016-02-26
feat(service): Introduce post processing functionality

tspaeth:
In some special cases, a developer or translation agency might need an easy way to see the contexts they're translating content for. 
To mark the keys and translations inside a page, the postProcessor gives the application developer the possibility to have such a behavior. It can also be used for other cases where translated content must be rewritten.

PR contains tests + new test case ex14_async_loader_with_postProcessor.htm

Open point: Shall we sanitize the output as this is intended to be an experimental / dev stage feature as of now?

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -41,6 +41,8 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
       loaderCache,
       directivePriority = 0,
       statefulFilter = true,
+      postProcessFn,
+      $postProcessEnabledState,
       uniformLanguageTagResolver = 'default',
       languageTagResolver = {
         'default': function (tag) {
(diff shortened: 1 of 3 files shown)


knalli:
@tspaeth  Does this duplicate a feature which `$http.postTransform` would already? https://github.com/angular-translate/angular-translate/commit/895268b734cc192b122692d4c0fa53971e73b4ed

tspaeth:
Hm, I'll have a closer look at that example (by the way: do we have some documentation on that somewhere? Couldn't find it quickly) - didn't know that. 
Maybe, some things of the PR could be achieved with this, but in the $http.transformResponse - case, the translation values itself need to be modified, right?

knalli:
The feature is indirect documentation via `$http` option in our loaders and the `transformResponse` by Angular itself. Eventually it is simply not so obvious.

> the translation values itself need to be modified, right?

Don't got it. You can change or modify the result in any way you want to do.

tspaeth:
Yes, you can do parts as far as I just tested. 
But the advantage of the later "intercepting" is that you can react on the language you're in as a target (in case of fallbacks - can you get this in the $http.transformResponse? -- I'm not sure) and you can work on interpolation results / interpolation mistakes. And it will work on all async / in-memory loaders & translation-key models as they're already normalized in that stage.

We've had a project request as in the above description for translation purposes to have a small "WYSIWYG" - feature (or more "What you see is where you need to translate the context" for translators.

src/service/translate.js
```diff
@@ -952,6 +954,50 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
...
+       *
+       * @param {boolean} sets the postProcess function to enabled or disabled (true / false) or a default false
+       */
+  this.postProcessEnable = function (flag) {
+    if (flag === undefined) {
+      $postProcessEnabledState = false;
+    } else {
+      $postProcessEnabledState = flag;
+    }
+    return this;
```
knalli: Why do we need tracking both states? If this could be changed at runtime I would understand.. but this this are one-time configuration things. Either there is a post processing function, or there is no one.

knalli:
I have remastered this a little bit, landed as fab03c9

* merge conflict in canary
* removal of the activator
* add option for service name

PR #1424, opened 2016-03-03
feat(service): introduce small function to return the registered avai…

### LINKED ISSUE #1304: Is there a getter for the registered available languages?
Hi,
I'm building an app which every time I need to do stuff with languages I'm making a request to my languages endpoint....

I'd like to register the available languages in the translate provider, and then use it instead of that request to my api/languages endpoint.

I couldn't find a way to get and use those registered languages in my controllers and directives, I'm mixing up stuff?

Thanks

tspaeth:
...lable languages also from the $translate - service and not only in the configuration phase - closes #1304

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -2115,6 +2115,25 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
         return deferred.promise;
       };
 
+      /**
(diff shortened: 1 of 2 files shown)


knalli:
Thanks @tspaeth , picked as 3b5b413

Issue #1413, opened 2016-02-24
Key is used as fallback instead of original text


iamdriz:
If I have a link like so:
`<li><a ng-class="{ active : $state.includes('section1') }" ui-sref="section1" translate="SECTION1">Introduction</a></li>`

And the translate key value `SECTION1` doesn't exist in the currently selected language it will instead use the fallback language key value instead which is great! However if it doesn't find the fallback language key value either, instead of just falling back to what the original text was, it just falls back to the key itself, in this case `SECTION1`. What I want is for it to just say `Introduction` if it doesn't find the `SECTION1` key value in either the selected or fallback language.

Is this a missing feature? Or a bug?

knalli:
Did you have already read our documentation and guide at https://angular-translate.github.io/ ?

You can either use a [custom error handler](https://angular-translate.github.io/docs/#/guide/17_custom-error-handler) or use the `default` extension of the [translate directive](https://angular-translate.github.io/docs/#/api/pascalprecht.translate.directive:translate).

I guess that is what you are looking for.

iamdriz:
Yeah read the documentation. However I'm not seeing anything that allows the current inline content to be used as the fallback. All the examples show either using a global error fallback or specifying the default as an attribute... I want it to use the actual inline content that is already there.

iamdriz:
@knalli Any updates on this? Thanks.

knalli:
http://plnkr.co/edit/tHWNfwDv0nTqhVWc08nx?p=preview

Alright, no, that is not possible. If you define a translation key, this will override the element's content (which will be used as an element id unless defined).

This is not a bug, perhaps a missing feature. If you want, you can make a pull request. :)

tspaeth:
@knalli didn't we have this request already in some other issues? I'll see what we can do ...

iamdriz:
@knalli Cool. Why do you believe this is not a bug out of curiosity? As if the key isn't defined in any translation, **why would you want the key to show and not the original content?**

tspaeth:
@iamdriz 
> why would you want the key to show and not the original content?

Well, at least speaking for the projects I am working on and where we use this module on, we're not putting dummy content in the node value/text, but only use Keys with either empty content/text or simply the key. 
And we want to see where texts are untranslated so we can feed in a translation at least for one language. To achieve that, we use an error handler but also we can see it on screen that it is untranslated.
I guess, the approaches are simply different we have in the wild.

iamdriz:
[...]

Issue #986, opened 2015-04-02
Directive should prefer "translate" attribute's value over element text


tiblu:
**Code**
`
  <pre translate="TRANSLATION_ID">This is the default</pre>
`

http://plnkr.co/edit/ZGqE7guPegfzEZ6YdJ6a?p=preview

**Result**

Translate "evaluation" order:
- "This is the default"
- "TRANSLATION_ID"

**Proposed**

Proposed "evaluation" order:
- "TRANSLATION_ID"
- "This is the default"

**Problem**

When using `angular-translate-handler-log` despite the fact that "MY_KEY" exist, warning is logged to "Translation for This is the default doesn't exist".

I find it useful to keep original text in the HTML element, so that template makes more sense when you look at it. I don't want to replace all the default texts with translation keys. It makes it harder to follow the template and merge any changes made by the HTML coders.

Plus the elements contents then acts as "default" text when no translation is found.

And the pure template can be opened and viewed in the browser as-is.

knalli:
Adding an option which do not refresh the directive's text until the translation has been successful (not rejected)? Okay..

However, one note: This also means further translation rejections will result into invalid states. Because no refresh would mean that an already applied translation will be not overridden (but is wrong). And even if there would be define an additional condition (like only on initial time = first success) the intended behavior would not be possible after the next language change.

I think that's okay, but it should be at least noteable.

darrylkuhn:
I wanted this too so I took a stab at it. You can see it here https://github.com/angular-translate/angular-translate/compare/master...aristamd:master and for those that just want to bring it into their project a complied / minified version is available here: https://github.com/aristamd/bower-angular-translate. 

Not sure all the implications of these changes since I'm new to the project. I changed a few of the test cases since the expected behavior had changed but otherwise all tests pass. Would love to merge this in to origin if suitable. Cheers!

soluml:
+1 ... I'm currently seeing the `TRANSLATION_ID` or a brief second before the translated text comes through. It's very noticeable on slower connections. Would love to default it to something else or hide the text entirely while it waits for the translation.

knalli:
@soluml it looks like a regular issue about FOUC which is handled in the guide. And this issue is not handling that kind of situation.

soluml:
Sorry, it sounded very similar to this issue. I've tried translate-cloak and it doesn't solve the issue. If I had the default text showing instead of the variable ID, it would look nicer.  I'll keep looking around the docs, thanks!

[...]

PR #1426, opened 2016-03-03
feat(directive): introduce new attribute translate-keep-content to no…

tspaeth:
…t replace the innerText/innerHTML of an element having a translation key
fix(service): fixes a bug being in the fallback stack that never rejected fallback promises

Might implement / fix issues from #986 / #1361 / #1413

### DIFF
--- a/test/unit/directive/translate.spec.js
+++ b/test/unit/directive/translate.spec.js
@@ -805,4 +805,51 @@ describe('pascalprecht.translate', function () {
       expect(element.attr('title')).toBe('Hallo');
     });
   });
+
+  describe('fallback, translateKeepContent and no missing translation handler', function () {
+
+    var $compile, $rootScope, element;
+
+    beforeEach(module('pascalprecht.translate', function ($translateProvider) {
+      $translateProvider
+        .translations('en', {
+          'HELLO': "Hello"
+        })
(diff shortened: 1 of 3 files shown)


test/unit/directive/translate.spec.js
```diff
@@ -805,4 +805,51 @@ describe('pascalprecht.translate', function () {
...
+    it('should use leave the inner content without change', function () {
+      element = $compile('<translate translate-keep-content>HELLONOTFOUND</translate>')($rootScope);
+      $rootScope.$digest();
+      expect(element.html()).toBe('HELLONOTFOUND');
+    });
+
+    it('should use leave the inner content without change if key provided as attribute', function () {
+      element = $compile('<translate="HELLO" translate-keep-content>My template content</translate>')($rootScope);
+      $rootScope.$digest();
+      expect(element.html()).toBe('My template content');
```
knalli: I don't understand why this should be not `HELLO`. It's a valid key.

tspaeth: Oh. This was just a reviewer test ;-). 
I'll have a look - you're certainly totally right.

knalli:
Picked as b2cf8a3

PR #1432, opened 2016-03-07
feat(service): improve messageformat.js output caching

eemeli:
Previously, the `$translateMessageFormatInterpolation` interpolator only cached results when both the
ICU MessageFormat string and the interpolation parameters matched. With this change, the compiled
function output by messageformat.js is cached, resulting in fewer cache misses that would require
the comparatively heavier `$mf.compile()` operation to be run.

An `mf:` namespace prefix is also added to the cache key, to make sure that a MessageFormat string like `'en'` would not match with the cached value of the `$mf` variable.


knalli:
@eemeli  Thank you!

knalli:
Thank you again. Picked as cb31608

Issue #1390, opened 2016-02-03
determinePreferredLanguage() + registerAvailableLanguageKeys() still try to load unavailable language


PaulAnnekov:
When using `determinePreferredLanguage()` function to automatically detect language based on Navigator and `registerAvailableLanguageKeys()` to set available languages list, if unavailable language detected, then angular translate still tries to load it using translation loader. Why does it do this? It's unavailable.
POC: [plunker](https://plnkr.co/edit/cm18nwAYmVcohYtWNDSe?p=preview)

knalli:
Yeah, this sounds valid.

We have to revisit https://github.com/angular-translate/angular-translate/blob/c5c20ac7e06d8569c79019b340102ce30fad8572/src/service/translate.js#L183-L185

tspaeth:
Question - Implementation should be:

No matter, what determinePreferredLanguage analyzed - the registerAvailableLanguageKeys - output defines the valid languages. 
Therefore if your browser is de-DE and this is the preferred due to this - which of the languages shall be chosen as the active one by the module?

Usually, you could now define this wildcard mapping as a second parameter (maybe needing the correct mappings in before - didn'T test it:
```
registerAvailableLanguageKeys(languages, {
      "*": "ar"
    })
```

knalli:
> Therefore if your browser is de-DE and this is the preferred due to this - which of the languages shall be chosen as the active one by the module?

The fallback one.

Perhaps we should log a warning if the combo preferredLanguage + available + fallback will result into nothing..

knalli:
Fixed with bde935e -- with test case.

Issue #1153, opened 2015-08-05
Unexpected behavior with fallback languages and the translate directive


proog:
I'm seeing some strange behavior with fallback languages and the translate directive. I made [a plunker here](http://plnkr.co/edit/yztOrpeAeSC795E3AbeC?p=preview) to demonstrate.

The intention is that there are three languages that each change certain terminology, and the fallback stack is changed during runtime.

It starts out fine, with the pirate language displaying "Ahoy world" (from the sailor language) and "Arrrrr". Clicking the Sailor button replaces the fallback stack and switches to the sailor language. The paragraph is changed accordingly, but the title is not -- instead, it is inexplicably changed to "Hello world", even though the sailor language has a translation for this key. Clicking Sailor again finally changes the title to the correct "Ahoy world".

It seems that using the translate filter instead of the directive fixes the issue. I can reproduce the problem reliably in both Chrome and Firefox. Any idea what's going on here?

tspaeth:
Please try to change the definition order in the change function

```
    $translate.use(lang);
    $scope.fallbacks = $translate.fallbackLanguage(fallbacks);
    $scope.lang = $translate.use();

```

This is one of the things, where we'll have to clean up - it's pending a while, but it'll take a while still... :-|

proog:
Thanks for your reply.

Your suggestion does fix my specific example, but introduces an issue where clicking Sailor and then Pirate changes "Ahoy world" to "Hello world", which it shouldn't.

proog:
So if I understand correctly, this is a known issue?

It seems to me that the options for working around this issue are to avoid changing fallbacks during runtime or use the translate filter instead of the directive.

knalli:
note: We will rework this in the next major version -- at least.

knalli:
Hello,

in part of a "Spring Cleaning 2017", we are closing this issue as part of several others. The meta discussion is available [here](https://github.com/angular-translate/angular-translate/issues/1671) for full transparency.

At the end, we are a small team and have only limited resources (both time and human) available. Said this, we must keep the focus and scope on the project.

We are grateful and appreciate to everyone for each made contribution (regardless being code, idea, documentation, and support).

We are deprecating/closing all issues which fulfills a specific set of criteria: mainly outdated, nice-to-have, or just out-of-scope.

----

Although the list has been curated to the best of our's knowledge, there will be some lack of understanding.

Okay, there a different stories behind all open issues, but we do not have the time for everything:

[...]

Issue #399, opened 2014-02-25
Using $translateProvider.useLoader, the addPart/refresh combination does not work properly for the frist element's translation.


virtualit786:
Hi Pascal,

Here is the config 
```js
.config(['$translateProvider', '$translatePartialLoaderProvider', function ($translateProvider, $translatePartialLoaderProvider) {

    $translateProvider.useLoader('$translatePartialLoader', {
        urlTemplate: 'i18n/{lang}/{part}.json'
    });
    $translateProvider.preferredLanguage('en-EN');
    $translatePartialLoaderProvider.addPart('main');
}])
```
Tranlsation files are :- 
i18n/en-EN/main.json
i18n/en-EN/pa.json
i18n/fr-FR/main.json
i18n/fr-FR/pa.json

The files have few common elements for example page heading is common in all files which refers to the heading of that particular page.

MainCtrl for mainview.html
```js
angular.module('appApp')//organization-definition
    .controller('MainCtrl', ['$scope', '$translate', '$translatePartialLoader',
        function ($scope,   $translate, $translatePartialLoader) {            
            $translatePartialLoader.addPart('main');
            $translate.refresh();
        }
    ]);
```
PaAddCtrl for paview.html
```js
angular.module('appApp')
    .controller('PaAddCtrl', ['$scope', '$translate', '$translatePartialLoader',
        function ($scope, $translate, $translatePartialLoader) {
            $translatePartialLoader.addPart('positions');
            $translate.refresh();
        }])
```
When the app starts, MainCtrl is the one so the language part sets up to main and it works fine. When I move to PaAddCtrl from MainCtrl, then the translation of first element does not come from the positions.json file but instead comes from the main.json file. 
And same thing happens when I move from PaAddCtrl's paview.html to MainCtrl's main.html. 

What am I doing wrong? Can you please point it out and help me fix this?

gtbuchanan:
Yes, this is still a problem. See [this][0] plunker taken from [this][1] answer on StackOverflow. I can confirm this is also still the case if I use the `$translate` service in the controller after `$translate.refresh` is called.

[0]: http://plnkr.co/edit/sZC2XST8BMZcMCYbgtxt?p=preview
[1]: http://stackoverflow.com/a/25070429/1409101

gtbuchanan:
This is my current undesirable workaround:

    app.run(['$rootScope', '$translate', '$interval', function ($rootScope, $translate, $interval) {
[...]

PR #1433, opened 2016-03-07
fix(service): fix and cover edge case where fallback language is set …

### LINKED ISSUE #1368: Fallback language translations failing to be processed.
I have angular translate configured using the following code.

```javascript 
        $translateProvider.useStaticFilesLoader({
            files: [{
                prefix: 'i18n/sys-captions.',
                suffix: '.json'
            }]
        }).fallbackLanguage('en_US')
        .determinePreferredLanguage();
```

In this instance determinePreferredLanguage correctly sets my preferred language as en_GB (I'm currently experimenting with how reliable it is). I only have the US translation file available and the fallbackLanguage file is correctly loaded from the server.

However, when I try and translate an item via the $translate service, it will fail to find the item. The problem seems to reside inside the determineTranslation function.

```javascript
if ($uses && $fallbackLanguage && $fallbackLanguage.length) {
   fallbackTranslation(translationId, interpolateParams, Interpolator, defaultTranslationText)
```

In my cases the $uses variable appears as undefined and so it does not make an attempt to process the fallback translation. Removing the $uses variable from the if statement corrects the issue for me.

tspaeth:
…and preferred language is determined. but use is not explicitly called before calling the $translate service

Fixes #1368 

plus possibly #1390, #1153 -- maybe #399?
I didn't have time to test all of these.

Fix is a proposal - so @knalli have a look if this is the best place to "fix" this or if we should do this in the determinePreferred already before, but then we have an order dependency on the initializers still.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1021,7 +1021,9 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
           startFallbackIteration;
 
       var $translate = function (translationId, interpolateParams, interpolationId, defaultTranslationText, forceLanguage) {
-
+        if (!$uses && $preferredLanguage) {
+          $uses = $preferredLanguage;
+        }
         var uses = (forceLanguage && forceLanguage !== $uses) ? // we don't want to re-negotiate $uses
               (negotiateLocale(forceLanguage) || forceLanguage) : $uses;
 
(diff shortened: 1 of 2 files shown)


knalli:
Alright @tspaeth I've picked as 02688f2 (w/ modified text).

PR #1459, opened 2016-04-11
fix(service): translate-language not working first time

Rafvb:
We had a problem with the translate-language directive. When we used it with to force a different language on a child scope, the first time it still showed the preffered language. When we opened the page again, it showed the correct translations.
I debugged and found that the promise that is selected as the promiseToWaitFor is based on the prefferedLanguage and not the forced language. The code change fixes this, but I'm not entirely sure that it didn't break anything else.
Also note that the demo in the translate language directive api reference (or the plunker) does not seem to work (https://angular-translate.github.io/docs/#/api/pascalprecht.translate.directive:translateLanguage).

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1086,9 +1086,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
         }
 
         var promiseToWaitFor = (function () {
-          var promise = $preferredLanguage ?
-            langPromises[$preferredLanguage] :
-            langPromises[uses];
+          var promise = langPromises[uses];
 
           fallbackIndex = 0;
 


knalli:
Sidenote: we have already fixed the typo in the docs resulting into an error.

knalli:
@Rafvb Can you provide a full working demo or (better) add an unit test covering our specific use case?

This issue here is still open, because I'm also in doubt this is breaking. Yes, no tests are broken but the promise waiter should wait for the preferred language also.

I'll not close this here, but your additional input would speed up the process. Thank you very much!

Rafvb:
@knalli Indeed, I will try to add a unit test.
Its a bit busy at the moment, so I'm not sure if I can add it this month.

I'm also not really sure this did not break anything else.

Rafvb:
Hi @knalli 
This seems to be fixed in the latest release.
When I use the latest release in the plunker example it works:
[http://plnkr.co/edit/7aK8JaIGvQklgkuEMbgt?p=preview](http://plnkr.co/edit/7aK8JaIGvQklgkuEMbgt?p=preview)

But when I replace the same code with an older version of ng-translate (2.8.0) it does not work:
[http://plnkr.co/edit/rVdOQLnzoIN8jCMbFXHf?p=preview](http://plnkr.co/edit/rVdOQLnzoIN8jCMbFXHf?p=preview)

So the problem was fixed somehow 👍

knalli:
OK - I understand this as being not reproducible any more.

Issue #1467, opened 2016-04-20
Allow multiple parts in translate value key translate-value-*


morgoth:
### Subject of the issue
Currently when using `translate` directive with passed value, like `p(translate="KEY", translate-value-something)` it's only possible that `something` is a single word.
This doesn't work: `translate-value-camelCased`, `translate-value-underscored_key`, `translate-value-hyphened-key`
I think it would be useful to support, if not all at least some of this syntax.

### Your environment
* version of angular-translate - 2.11.0
* version of angular - 1.5.5
* which browser and its version - chrome 49

knalli:
What do you mean?

```html
<span translate="key" translate-value-my-special-thing="123"></span>
```

where you key would be `my-special-thing`? If so: No, that is not possible. The shortcut depends on the way AJS reads and transforms the attribute names. You have to use `translate-values` for more complex or invalid values.

morgoth:
Yes, I was thinking exactly about such keys.

Actually, when I now look at the source https://github.com/angular-translate/angular-translate/blob/adeaa3e3a19f46cc4d40de8665f7e4492fce8475/src/directive/translate.js#L141

it works, but result is rather surprising, having key `translate-value-my-special-think` it is available as `mySpecialThink` - do you think it could be documented?
I think it would be more clear if the key would be "underscorized" to `my-special-think`, but this might be a breaking change.

What do you think?

knalli:
No, that is the way AngularJS is handling element attributes. Just like directive names or directive scope bindings. We have not reinvented the wheel for this feature.

Feel free to send a helpful addition into the guide, but it is not a surprise.

PR #1472, opened 2016-04-20
Add example of camel cased interpolation key in directive

morgoth:
Follow up to https://github.com/angular-translate/angular-translate/issues/1467


knalli:
Thank you, picked as 5dc6e78

PR #1481, opened 2016-04-28
fix(grunt): Work around uglify to preserveComments as expected

### LINKED ISSUE #1461: Strip comments from .min.js
### Subject of the issue
It looks like at least latest version includes all jsdoc comments in minified version https://cdn.jsdelivr.net/angular.translate/2.11.0/angular-translate.min.js.

### Expected behaviour
Doc comments must be stripped from minified version.

marclr:
fix(grunt): Work around uglify to preserveComments as expected until it's fixed in their repo
Root cause: gruntjs/grunt-contrib-uglify#366 
Fixes #1461

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -282,7 +282,7 @@ module.exports = function (grunt) {
     uglify: {
 
       options: {
-        preserveComments: 'some'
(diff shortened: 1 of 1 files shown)


knalli:
@marclr : Thank you, picked as 32cdedb

PR #1485, opened 2016-05-03
fix(service): prevent capitalization confusion

austinjalexander:
Greetings!

---

Given a key registration:
```
  .registerAvailableLanguageKeys(['en-us', 'zh-cn'], {
    'en*': 'en-us',
    'zh*': 'zh-cn',
    '*': 'en-us'
  })
```

We noticed an example where `en-US` was not being mapped to `en-us` as we expected.

This issue seems to be caused by:
```
if (indexOf(avail, locale) > -1) {
  return preferred;
}
```
Using our example, this bit would check whether or not the _lowercase version_ of 'en-US' is a registered language, which it is (in the form of 'en-us'), but then the mixed-cased version ('en-US') is returned instead of the registered key.

---

Thank you for `angular-translate`!

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -159,9 +159,9 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
       avail.push(angular.lowercase($availableLanguageKeys[i]));
     }
 
-    // Check for an exact match in our list of available keys
+    // Check for an exact (lowercase) match in our list of available keys
     if (indexOf(avail, locale) > -1) {
-      return preferred;
+      return $availableLanguageKeys[indexOf(avail, locale)];
     }
 
     if ($languageKeyAliases) {
--- a/test/unit/service/translate.spec.js
+++ b/test/unit/service/translate.spec.js
@@ -92,8 +92,9 @@ describe('pascalprecht.translate', function () {
 
     beforeEach(module('pascalprecht.translate', function ($translateProvider) {
       $translateProvider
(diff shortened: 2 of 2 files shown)


src/service/translate.js
```diff
@@ -159,9 +159,9 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
       avail.push(angular.lowercase($availableLanguageKeys[i]));
     }
 
-    // Check for an exact match in our list of available keys
+    // Check for an exact (lowercase) match in our list of available keys
     if (indexOf(avail, locale) > -1) {
-      return preferred;
+      return $availableLanguageKeys[indexOf(avail, locale)];
```
tspaeth: Just saying: a 
```
return locale;
```
would also do, right?

knalli:
@austinjalexander Before going deeper into this PR: Are you aware of [`$translate.uniformLanguageTag()`](https://github.com/angular-translate/angular-translate/blob/adeaa3e/src/service/translate.js#L824) ?

knalli:
Closed due inactivity.

In general, the feature "uniform language tag" is proposed to have an internally general outline of languages. Because of different environments, language code styles/standards.

If this really does not solve the problem, we are welcome any (new) contribution. But until this happens, this is outdated here.

austinjalexander:
@knalli Thank you! I'll check to see if `$translate.uniformLanguageTag()` does the trick!  :)

PR #1499, opened 2016-05-23
fix(service): allow instant function to also take care of post proces…

tspaeth:
…s configuration

fixes an issue where the newly introduced postprocessFn was not working with $translate.instant() - this fix now also corrects the behavior for this corner case.
But remember: instant - calls will not take care of async loaders!


knalli:
tldr: Does this fixing open issues?

tspaeth:
I've not seen any issue yet coming up besides that it popped up in the customer project where the "instant labels" were missing the post processing information.

knalli:
Ok, picked

PR #1502, opened 2016-05-27
feat(directive): introduce a global keepContent setting

timsly:
Introduce a way to set `keepContent` globally.

If we set it to true we don't need to add `translate-keep-content` to every directive, it will become a default behavior


knalli:
Thank you, great addition. 🎉 

Landed as 2015f79

PR #1507, opened 2016-05-31
docs(guide): Information on post process function in error handling

### LINKED ISSUE #1501: browser language detection causes app to error when user is not using a supported language
### Subject of the issue
When using 
`$translateProvider.useLoader('$translatePartialLoader' ...`
with 
```
$translateProvider.registerAvailableLanguageKeys(['en', 'ja'], {
    'en*': 'en',
    'ja*': 'ja'
  }).determinePreferredLanguage();
```
and
`$translateProvider.fallbackLanguage('en');`

Angular-translate still looks for language files that are not registered. 
Is this the expected result/behavior?

### Your environment
* version of angular-translate - 2.11.0
* version of angular 1.5.5
* which browser and its version - IE11, Chrome Version 49.0.2623.87 m

### Steps to reproduce
Set your preferred browser language to one other than 'en' or 'ja' (these two are available in the application, I used 'es')
Run this plnkr [(https://plnkr.co/edit/VeeJSTK1xKUNSkv7CzVz?p=preview)]

### Expected behaviour
The app should load the fallback language when the user's preferred language is not registered

### Actual behaviour
The application errors because it is looking for a <lang>.json that doesn't exist. It doesn't even load the fallback language.

tspaeth:
docs(guide): Improve fallback documentation on wildcard

Closes #1501 documentation request
Adds extended documentation for #1421 feature

Languages provided within this PR:
- en
- de

### DIFF
--- a/docs/content/guide/de/09_language-negotiation.ngdoc
+++ b/docs/content/guide/de/09_language-negotiation.ngdoc
@@ -76,9 +76,18 @@ $translateProvider
   .determinePreferredLanguage();
 </pre>
 
-Diese Funktionalität ergibt nur Sinn in Kombination mit
-'determinePreferredLanguage()'!
+Wenn man sicher gehen will, dass auch Sprachschlüssel, die nicht explizit erwähnt werden, auf eine der unterstützten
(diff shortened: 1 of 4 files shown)


knalli:
picked

Issue #1509, opened 2016-06-02
Help: Forcing another language in controller not working


SandeepThomas:
* version of angular-translate - 2.11.0
* version of angular - 1.5.5

Trying to get text translated in specific language irrespective of the current one. But always getting text in the current language only.

This is the code I'm using 
$translate('hello', null, null, 'de').then(function (translation) {
  _this.forcedtext = translation;
});

Please see this plunker: https://plnkr.co/edit/M48AZbQgonPctpO3r8nt?p=preview

Can someone please point out to me what I'm doing wrong? 

Thanks

tspaeth:
use 'de' as the 5th parameter, not the 4th.

SandeepThomas:
oh.. thank you..

i should use it like this right?

$translate('hello', null, null, null,'de').then(function (translation) {
	_this.forcedtext = translation;
});

tspaeth:
API-Documentation needs update here as far as I can see. :-)

knalli:
Corresponding pr closed already.

PR #1510, opened 2016-06-02
fix(service): Correct descriptive ngdocs to match parameters on the

tspaeth:
…service calls

fixes issue raised in #1509


knalli:
picked

Issue #1516, opened 2016-06-16
$translate on missing key does not return the key, but a failed promise


RichardBradley:
### Subject of the issue

The [documentation says](https://angular-translate.github.io/docs/#/guide/03_using-translate-service) "If [the translation id] doesn't exist it'll probably just return the translation id."

On that basis, I would expect a call like:

    $translate("MISSING_KEY").then(function(txt) { console.log(txt); })

to log "MISSING_KEY".

However, it does nothing at all, because the promise returned from $translate is completed with a failure.

The correct code would be:

    $translate("MISSING_KEY").then(function(txt) { console.log(txt); }, function(txt) { console.log(txt); })

as you can see in the "`it('should return translation id if translation doesn\'t exist'`" test at [translate.spec.js:276](https://github.com/angular-translate/angular-translate/blob/adeaa3e3a19f46cc4d40de8665f7e4492fce8475/test/unit/service/translate.spec.js#L276)

This means that the documentation is misleading and makes code that uses the service difficult to write correctly. 

For example, [this code in `ng-admin`](https://github.com/marmelab/ng-admin/blob/c9dc3efdba38f8953535f2b6da715693f97c19c1/src/javascripts/ng-admin/Crud/delete/DeleteController.js#L47) will return nothing at all if the "errorMessage" translation key is not found:

                $translate(errorMessage, {
                    status: error && error.status,
                    details: error && error.data && typeof error.data === 'object' ? JSON.stringify(error.data) : {}
                }).then(text => notification.log(text, { addnCls: 'humane-flatty-error' }));

... because the author did not anticipate that the promise returned by `$translate` could fail. (I will raise a separate bug with `ng-admin` to try to fix this.)

### Your environment
* version of angular-translate: 2.11.0
* version of angular: 1.4.8
* which browser and its version: n/a

### Steps to reproduce

Reproduced in the unit test linked above.

### Expected behaviour

The promise returned by `$translate("MISSING_KEY")` should complete with the value `"MISSING_KEY"`

### Actual behaviour

The promise returned by `$translate("MISSING_KEY")` does not complete, but fails with an error value of `"MISSING_KEY"`

knalli:
Well.

Reading it literally, you are right: It will not return (implicitly read) the requested translation id on the success callback.

[...]

PR #1523, opened 2016-06-28
docs(guide): #1516 clarify documentation for failure cases

RichardBradley:
### Description

This is a fix for the issues discussed on #1516, namely that the docs do not mention that users must include a failure callback when resolving the Promise returned by `$translate`, if they want to handle the case when the translation key was not found.

@knalli -- is this what you had in mind? I've done only the English version for now; I'll fix up the other languages if you are happy with this.

### DIFF
--- a/docs/content/guide/de/03_using-translate-service.ngdoc
+++ b/docs/content/guide/de/03_using-translate-service.ngdoc
@@ -40,12 +40,18 @@ eines Controllers wie folgt aussehen:
 app.controller('Ctrl', ['$scope', '$translate', function ($scope, $translate) {
   $translate('HEADLINE').then(function (headline) {
     $scope.headline = headline;
+  }, function (error) {
+    $scope.headline = error;
   });
   $translate('PARAGRAPH').then(function (paragraph) {
     $scope.paragraph = paragraph;
+  }, function (error) {
+    $scope.paragraph = error;
   });
   $translate('NAMESPACE.PARAGRAPH').then(function (anotherOne) {
     $scope.namespaced_paragraph = anotherOne;
+  }, function (error) {
+    $scope.namespaced_paragraph = error;
   });
 }]);
 </pre>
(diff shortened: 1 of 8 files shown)


knalli:
👍  Yeah, I think that information you had to want to see at least. ?

Perhaps with an addition marker that error is the requested translationId (key).

tspaeth:
LGTM - I also share knalli's remark :-)

RichardBradley:
> Perhaps with an addition marker that error is the requested translationId (key).

Please could you clarify what you mean here? I don't quite follow, sorry.
Maybe comment on which line you'd like me to change in the "files changed" view?

RichardBradley:
I have made the same changes to the code in all the translated docs and updated this PR.

I have not been able to copy across the wording changes to the translations, as I don't speak those languages. :-(

knalli:
> I have made the same changes to the code in all the translated docs and updated this PR.

It's all fine!

> Please could you clarify what you mean here? I don't quite follow, sorry.

You have named the argument `error`, but as you already pointed out yourself this will be misleading. It's actually not an error, but a translationId. The docs should care about this.

RichardBradley:
> It's actually not an error, but a translationId.

Well, it will be the translationId in the case where the translation Promise has failed due to a key-not-found.
If the Promise has failed for any other reason, it will be something else.

As discussed on #1516, I think it's confusing and potentially ambiguous to use the Promise "error" value to return the translation Id here.

I think user code should either look like 1:

```
$translate("MY_TEXT").then(function(translation) {
  stringFromService = translation;
}, function(error) {
  stringFromService = error;
});
```

or 2:

```
var stringFromService;
var translationId = "MY_TEXT";
$translate(translationId).then(function(translation) {
  stringFromService = translation;
}, function(error) {
  // Unable to translate given translationId; fall back to displaying the ID to the user:
  stringFromService = translationId;
});
```

I have used (1) in the docs in the attached PR, as (2) is quite verbose.

OK to leave, or do you strongly prefer 3:

```
$translate("MY_TEXT").then(function(translation) {
  stringFromService = translation;
}, function(translationId) {
  stringFromService = translationId;
});
```
?

knalli:
Sorry for the delay. Yes, I prefer 3. You are right this is not what it should be -- but that is how it *is* at the moment. There are several issues in the project which could be fixed -- but that would require a major release because significant API changes. I'm always trying not creating breaking changes except major releases -- that's semver.

knalli:
I'll care of this.

knalli:
picked as c0b0809

RichardBradley:
Thanks, looks good to me

PR #1530, opened 2016-07-04
Fix issue with 'escapeParameters' strategy when using jQuery and obje…

### LINKED ISSUE #1529: Issue with "escapeParameters" sanitization strategy and jQuery.
Hi, 

I'm having an issue when using "escapeParameters" sanitizing strategy and "jQuery".
Actually, after some debug, I found that when passing a object as an interpolation parameter it gets processed by `mapInterpolationParameters` and calls `htmlEscapeValue` on every property including functions.
The problem is that the current escaping function calls `angular.element('<div>').text(value)` but when using `jQuery` if value is a function, it will be called by `element.text`.

An additional condition should be added to check that if a value is a function then it should not be escaped.

By the way, it could be better to contribute to "MessageFormat" and allow it to only escape the properties when used in the message format otherwise with the current approach, each translation escapes all the properties even if they are never used.

Here's the associated plunker.
https://plnkr.co/edit/NvpKzg?p=preview

Thank you in advance!

yjaaidi:

### Description

💔Thank you!

…cts containing functions (not methods).

Fixed #1529

### DIFF
--- a/src/service/sanitization.js
+++ b/src/service/sanitization.js
@@ -253,6 +253,12 @@ function $translateSanitizationProvider () {
 
       stack.push(value);
       angular.forEach(value, function (propertyValue, propertyKey) {
+
+        /* Skipping function properties. */
+        if (angular.isFunction(propertyValue)) {
+          return;
+        }
+
         result[propertyKey] = mapInterpolationParameters(propertyValue, iteratee, stack);
       });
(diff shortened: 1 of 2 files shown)


knalli:
Picked as 492d8e5

I've added data to the mocking functions, as this fails on AJS 1.5

Issue #1297, opened 2015-11-17
Disable SanitizeValueStrategy for one $translate call


guillaume-fr:
I configured `translateProvider` with `'sanitize'` strategy and want to keep it to prevent injection.

There is a chart in my application draw using a canvas, I want to disable (or change) the `SanitizeValueStrategy` when I call `$translate()` to retrieve text (series names) that will be written on the canvas.

I don't know if this is already possible, otherwise it could be great to have a parameter for SanitizeStrategy or a dedicated function like `$translate.unsafe()` ignoring sanitize.

mwz:
Having ```$translate.instant()``` accepting a null sanitization strategy would be useful, alternatively calling ```$translateSanitization.sanitize(value, mode, "null")``` would also work for me.

I've been using the following as a workaround:

    $translateSanitization.useStrategy(null);
    var translation = $translate.instant('KEY');
    $translateSanitization.useStrategy('escape');

RenaldasK:
+1 Having same issue

knalli:
Ok. I have tagged this as "Contribution Welcome". (hint-hint).

PR #1546, opened 2016-07-28
Feat translate instant sanitize strategy

spiroid:
## Motivation

Sometimes non html messages need to be translated. For example in notification popups, etc.
The only solution as for now, is to the disable sanitize strategy globally

```javascript
$translateProvider.useSanitizeValueStrategy(null);
```

or to use this kind of hacks:

```javascript
$translateSanitization.useStrategy(null);
var translation = $translate.instant('KEY');
$translateSanitization.useStrategy('escape');
```

which disables sanitization globally for a short period of time, and does a call to $translate.instant in that time lap.

## Proposed solution 

This modification add a `sanitizeStrategy` parameter to the $translate.instant function to allow to override sanitization strategy for one call. Without having to change it in the service configuration for all `$translate` calls.

### Usage

To disable sanitization: 

```javascript
var translation = $translate.instant('KEY  {{ value }}', { value: '<p>hello</p>' }, null, null, null);
console.log(translation)

-> KEY <p>hello</p>
```

The important parameter is the 5th one, and it could take any valid sanitization strategy: escape, escapeParameters, etc.  Including the `null` (!= undefined) value which disables sanitization.
   
Relates #1297

### DIFF
--- a/docs/content/guide/de/15_custom-interpolators.ngdoc
+++ b/docs/content/guide/de/15_custom-interpolators.ngdoc
@@ -20,7 +20,7 @@ die ein Objekt zurück gibt, die folgendes Interface implementiert:
 
 - `setLocale(langKey)` - Setzt die Sprache mit Sprachschlüssel
 - `getInterpolationIdentifier()` - Gibt einen Identifier für den Service zurück
-- `interpolate(string, interpolateParams)` - Interpoliert Strings gegen Interpolationparamter
+- `interpolate(string, interpolateParams, sanitizeStrategy)` - Interpoliert Strings gegen Interpolationparamter
 
 So könnte ein custom Interpolation-Service aussehen:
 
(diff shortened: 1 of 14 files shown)


spiroid:
I can't reproduce the travis output.

Unit tests are running fine on my laptop. Moreover, it seems that the failing unit test has nothing to do with the refactored code.

spiroid:
hi @knalli,

Thanks for the feedback.
I updated  the PR to the latest commit on canary as of today.

I saw the context parameter :+1: 
It will surely help to have more context on what called the translation functions.

I assume it is better to use one of the new sanitize strategy than a null value. To be compatible with the existing code base, i kept that possibility.

knalli:
Thanks, looks good (fast read).

knalli:
Merged, thank you for contribution.

PR #1564, opened 2016-08-24
fix(service): fix infinite loop when fallback language async loading…

### LINKED ISSUE #1468: Infinite loop when fallback language async loading fails
### Subject of the issue
If angular-translate fails to load your preferred or fallback language files then it ends up in an infinite loop.

We’re experiencing this when our app auto-redirects the user to an external login service. The browser cancels the pending http request for the language file - and then the browser effectively hangs. 

Occurs on 2.11.0 but we first started having issues after this commit https://github.com/angular-translate/angular-translate/commit/40ad523.

This line in the promise error handler ends up recursively calling use() as the promise is already rejected.
```
return $translate.use($fallbackLanguage[0]).then(deferred.resolve, deferred.reject);
```

### Your environment
* version of angular-translate - 2.11.0 (occurs since 2.9.2)
* version of angular - 1.4.8 (also tried 1.5.5)
* which browser and its version - all I've tried

### Steps to reproduce
You can easily recreate it by attempting to load a non-existent language - e.g. with this plunkr by changing the fallback language to something that won't load - e.g. ```$translateProvider.fallbackLanguage("enX");```. Be prepared for you browser to hang when you do this.

https://plnkr.co/edit/hdP1U8?p=preview

### Expected behaviour
Should behave as if translation doesn't exist - i.e. the same as if you don't specify a fallback language and the async load fails. The rendered page might not look pretty but in our case the user never sees the page anyway because of the redirect.

### Actual behaviour
Browser hangs

nhenel:
Fixes #1468

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1910,7 +1910,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
             return translation;
           }, function (key) {
             // find first available fallback language if that request has failed
-            if (!$uses && $fallbackLanguage && $fallbackLanguage.length > 0) {
+            if (!$uses && $fallbackLanguage && $fallbackLanguage.length > 0 && $fallbackLanguage[0] !== key) {
               return $translate.use($fallbackLanguage[0]).then(deferred.resolve, deferred.reject);
             } else {
               return deferred.reject(key);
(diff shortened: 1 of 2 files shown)


knalli:
Thank you, we will have a look at asap.

eliasson:
Would it be possible to have a separate release including this fix?

knalli:
Yeah, ongoing..

knalli:
Picked as 233f30c

Thank you!

Issue #1374, opened 2016-01-23
Feature Request: Access list of translation items from external


iJungleboy:
Background
1. I have a fairly large AngularJS + angular-formly + angular-translate front end...
2. which uses 2-3 non-angular components like tinyMCE
3. I have custom translations on these third party, non-angular components
4. ...which I must provide with a list of translations while they load

I have two options: 
1. create another, completely separate language management for these components
2. manage the languages in angular-translate, and then write a few lines of code to pass on the relevant values

I prefer option two, but this requires that I can say something like (pseudo code)

```
var x = translate.dictionary.Extensions.TinyMce;
foreach(key in x)
   addToTinyMce(key, ...)

```

this is not possible. As far as i can see the list of translations cannot be retrieved from outside.

knalli:
@iJungleboy  Well, we can provide something like `$translate.getCurrentTranslations()` only as the `translate service` does not know  which translation could be possible. Said this, if you would call `getCurrentTranslations()` before the loader has (successfully) completed, you would even get an empty list. If you use something like a partial loader, you will get incomplete list. If you use a mixed setup with fallbacks, this actual "list of current translations" is not that easy as it looks in the first view.

I'll mark this as a feature request with call for contribution. A nice idea, but definitely with some work to do.

iJungleboy:
For anybody else trying to find a temporary workaround, my solution is to keep a comma-separated list of keys for all items I need to forward to TinyMCE, and then loop through these to lookup the translations and then forward them to TinyMCE. Not perfect, but works and keeps all my translations in one place.

tomlandia:
+1 would be perfect if you could read and overwrite translations from outside of angular. Right now I'm using a custom loader using translations from an external variable.

headcrabmeat:
How about if it just returns object like `$translationTable[traslation_id]`

knalli:
Landed.

iJungleboy:
Very cool :). I'll test it within the next few weeks.

iJungleboy:
To anybody who is interested - here's how I used the angular-translate table to translate TinyMCE in [2sxc](https://github.com/2sic/2sxc)

```javascript
    // Initialize the tinymce resources which we translate ourselves
    function initLangResources(editor, language, $translate) {
        var keys = [], mceTranslations = {}, prefix = "Extension.TinyMce.", pLen = prefix.length, primaryLan = "en";

        // find all relevant keys by querying the primary language
[...]

PR #1578, opened 2016-09-07
feat(service): add public method to get translationTable by id

headcrabmeat:
Feature Request  Access list of translation items from external #1374
* Implemented public method `$transate.getTranslationTable` in service
* Added UTs.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -2299,17 +2299,18 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
        * @description
        * Returns translation table by a key of the locale
        * if key is not provided it returns translation table which currently
-       * in use
+       * in use. If translation dictionary is in process of dowloading it returns
+       * null.
        *
        * @param {string} translationId A token which represents a translation id
        *
-       * @return {object} angular-translate $translationTable
+       * @return {object} a copy of angular-translate $translationTable
        */
       $translate.getTranslationTable = function (translationId) {
         if(!translationId) {
-          return $translationTable[$translate.use()];
+          return angular.copy($translationTable[$translate.use()]);
         }
-        return $translationTable[translationId];
+        return angular.copy($translationTable[translationId]);
       };
 
       // Whenever $translateReady is being fired, this will ensure the state of $isReady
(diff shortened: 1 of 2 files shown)


src/service/translate.js
```diff
@@ -2291,6 +2291,27 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
...
+       *
+       * @param {string} translationId A token which represents a translation id
+       *
+       * @return {object} angular-translate $translationTable
+       */
+      $translate.getTranslationTable = function (translationId) {
+        if(!translationId) {
+          return $translationTable[$translate.use()];
+        }
+        return $translationTable[translationId];
```
knalli: This expose the internal structure. This lets others modify *our* internal storage, but without knowing it (therefor no implicit update/refresh). This can/will result into confusion later.

test/unit/service/translate.spec.js
```diff
@@ -2740,6 +2744,41 @@ describe('pascalprecht.translate', function () {
...
+    var $translate, $rootScope, $STORAGE_KEY, $q;
+
+    beforeEach(inject(function (_$translate_, _$rootScope_, _$STORAGE_KEY_, _$q_) {
+      $translate = _$translate_;
+      $rootScope = _$rootScope_;
+      $STORAGE_KEY = _$STORAGE_KEY_;
+      $q = _$q_;
+    }));
+
+    it('should return transaltion table which is curently in use if no params', function () {
```
knalli: transaltion -> translation

src/service/translate.js
headcrabmeat: Good point. I like your idea to use `angular.copy()`.

knalli: Perhaps we should also add a test + doc about the situation when translation are loaded async but `$translate.getTranslationTable(lang)` will return null (i.e. because of loading in progress).

headcrabmeat: what if access it via promise model?

knalli: Wait for what? There is no trigger available.

headcrabmeat: Yep, my bad.

knalli:
Thank you. I've reworked this because some conflicted changes and minor styling issues.

Thank you.

PR #1593, opened 2016-09-21
Loader static file file map

JacobJohansen:

### Description
Adds a file map look up option to loader static files prevent browser cache issue.

The file map allows for the adding of revisioning numbers to the file to prevent browser caching issues.

💔Thank you!

### DIFF
--- a/src/service/loader-static-files.js
+++ b/src/service/loader-static-files.js
@@ -36,12 +36,18 @@ function $translateStaticFilesLoader($q, $http) {
         throw new Error('Couldn\'t load static file, no prefix or suffix specified!');
       }
 
+      var fileUrl = [
+        file.prefix,
+        options.key,
+        file.suffix
+      ].join('');
+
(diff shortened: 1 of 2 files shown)


knalli:
@JacobJohansen  Thank you.

Well, one question: If you can control this map, why you cannot control the requested file path? :)

Anyway. Im with this, but we should add this feature to the url loader also. 

Additionally, what do you think about adding an optional function resolver being more flexible?

JacobJohansen:
@knalli 
We the map contains a file name with a hash attached. its unique per file

JacobJohansen:
@knalli 
Looking at the URL loader, I'm not sure what would be gained by adding the file mapper feature since it's based on directly injecting the correct URL. it could still be added. 

I'm not sure how the function resolver would be more beneficial, I currently can't come up with any other use cases to justify adding the layer of abstraction.

knalli:
Yeah, you are right.

knalli:
Thank you.

Picked as 132e49a

PR #1594, opened 2016-09-22
fix(service): return empty string when found in fallback

nizhu:

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!

### Description
For some strings I'd like the default to be an empty string. Omitting the key/value pair `{}` and having a empty string value `{ EMPTY_DEFAULT: '' }` are treated the same - returning the KEY surrounded by not found indicators.

This change makes the behaviour of finding the fallback translation match that of finding a normal translation. See https://github.com/angular-translate/angular-translate/blob/master/src/service/translate.js#L2175

I've not looked in the angular-translate codebase before but I think this is a breaking change when the users have multiple fallback languages and an earlier one has an empty string (look at unit test - existing behaviour is returning `bare`)

💔Thank you!

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1507,7 +1507,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
         if (fallbackLanguageIndex < $fallbackLanguage.length) {
           var langKey = $fallbackLanguage[fallbackLanguageIndex];
           result = getFallbackTranslationInstant(langKey, translationId, interpolateParams, Interpolator);
-          if (!result) {
+          if (!result && result !== '') {
             result = resolveForFallbackLanguageInstant(fallbackLanguageIndex + 1, translationId, interpolateParams, Interpolator);
           }
         }
--- a/test/unit/service/translate.spec.js
+++ b/test/unit/service/translate.spec.js
@@ -2316,10 +2316,14 @@ describe('pascalprecht.translate', function () {
         .useLoader('customLoader')
         .translations('en', {
           'FOO': 'bar',
-          'BAR': 'foo'
+          'BAR': 'foo',
+          'BARE': ''
         })
         .translations('de', {
           'FOO2': 'bar2'
+        })
+        .translations('fr', {
+          'BARE': 'bare'
         })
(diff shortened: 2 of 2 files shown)


knalli:
> I've not looked in the angular-translate codebase before but I think this is a breaking change when the users have multiple fallback languages

Yeah, well. I'm seeing this from this point: If a translation is available explicitly (even an empty string), it should be used and not skipped nor ignored.

The improvement fixes also a buggy situation.

src/service/translate.js
```diff
@@ -1507,7 +1507,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
         if (fallbackLanguageIndex < $fallbackLanguage.length) {
           var langKey = $fallbackLanguage[fallbackLanguageIndex];
           result = getFallbackTranslationInstant(langKey, translationId, interpolateParams, Interpolator);
-          if (!result) {
+          if (!result && result !== '') {
```
FagnerMartinsBrack: Are we expecting only `undefined` here? If so can't we just `if (result === undefined)`?

nizhu: We're dealing with strings so `0`, `undefined`, `null`, `false` and `NaN` are all invalid IMO.

FagnerMartinsBrack:
@knalli Any news on this one?

knalli:
Ongoing. Stay tuned.

tspaeth:
In general LGTM. But from a readability view, I'd write it differently or just put this check in a function named "isDefinedOrEmpty" or similar.

FagnerMartinsBrack:
@tspaeth 

tbh, given the state of the codebase and that we have a decent coverage of this use case, I don't think this will add much value. Although I definitely agree that it is useful to create code composed of small functions because it can be self-documented.

I would suggest to add this issue as a comment to the test and create a function `!isValidFallback(fallbackValue)` (as you see, also renaming `result` to a more meaningful name since we are still working on this line).

knalli:
It's okay so far. Picked as d76227e

knalli:
Thank you for contribution.

PR #1624, opened 2016-10-22
Italian localization

markainick:
Added italian files, the docs still need translation though.


knalli:
Hello,

in part of a "Spring Cleaning 2017", we are closing this issue as part of several others. The meta discussion is available [here](https://github.com/angular-translate/angular-translate/issues/1671) for full transparency.

At the end, we are a small team and have only limited resources (both time and human) available. Said this, we must keep the focus and scope on the project.

We are grateful and appreciate to everyone for each made contribution (regardless being code, idea, documentation, and support).

We are deprecating/closing all issues which fulfills a specific set of criteria: mainly outdated, nice-to-have, or just out-of-scope.

----

Although the list has been curated to the best of our's knowledge, there will be some lack of understanding.

Okay, there a different stories behind all open issues, but we do not have the time for everything:

* Support: Because we have only limited resources, we cannot handle support questions any more. Please use appropriate boards like [Stackoverflow](http://stackoverflow.com/questions/tagged/angular-translate).
* Feature requests or ehancements: Altough it is probalbly a great idea (otherwise it would be closed already), we can not giving more resources to things out of scope or which are useful for too specific scenarios only.
* Bugs: Filed a long time ago, this bug either has a reasonable workaround, no effective solution without harming all others, or is lacking additional information/use-case.
* Docs: All documentation efforts not matching English will be abonded.

Thank you.

*angular-translate team*

----

# FAQ

**"But this is really important!"**

Well, we are all humans. Maybe we have overseen an important issue. If we missed something, we hope you are reaching as again *filing a new issue* with the reference to the old one. But please be aware: A bug without further details and a demo included will be rejected again.

**"But this would a great feature!"**

Reconsider: Are we really required to add something in order to make this feature happen? Or could you also create a project on your own utilizing this project?

PR #1640, opened 2016-11-02
Angular translate handler log strict DI support

tiblu:
Add strict DI support to Angular translate handler log (src/service/handler-log.js).

### DIFF
--- a/src/service/handler-log.js
+++ b/src/service/handler-log.js
@@ -11,7 +11,7 @@ angular.module('pascalprecht.translate')
  *
  * @returns {function} Handler function
  */
-.factory('$translateMissingTranslationHandlerLog', $translateMissingTranslationHandlerLog);
+.factory('$translateMissingTranslationHandlerLog', ['$log', $translateMissingTranslationHandlerLog]);
 
 function $translateMissingTranslationHandlerLog ($log) {
 


knalli:
We do have explicit dependency naming, see in artifact https://github.com/angular-translate/bower-angular-translate-handler-log/blob/2.13.0/angular-translate-handler-log.js#L22

tiblu:
It's there in the bower, but not in the source - https://github.com/angular-translate/angular-translate/blob/master/src/service/handler-log.js.

- How does it get added to the bower version?
- What's the consideration not having it straight up in the source code?

knalli:
We apply additional code within the build process.

tiblu:
Why in the build process, why not code it like you mean it? What's the benefit? Just out of curiosity.

knalli:
Because it is not relevant for the code to run. It's an external request for minified code and eventually the strict di mode.

The source code is for maintenance and readability, but not for the build result.

There is no benefit including these optional code right into the source code -- unless the origin source code files would be released. And that's is not the case.

tiblu:
OK, cool. Thanks for the insight and maintaining this great project!

knalli:
Anyway: Thank you for sharing contribution ideas.

PR #1650, opened 2016-11-16
fix(staticFilesLoader): do not use empty string as $http params

### LINKED ISSUE #1646: Incorrect value of 'params' in static loader request
According to angular docs 'params' should be a map, not string
https://docs.angularjs.org/api/ng/service/$http#usage

https://github.com/angular-translate/angular-translate/blob/132e49a62053491ae661f1796d8a509f7d902482/src/service/loader-static-files.js#L52

zgmnkv:
Closes #1646


tspaeth:
LGTM

knalli:
Thank you for your contribution.

Picked as ac2a038

PR #1651, opened 2016-11-16
feat(service): add translationId to custom interpolation service

joelongstreet:
### Included in this PR

* Adds `translationId` as a parameter to the method `interpolate` within the custom interpolation interface.
* Docs updated to reflect new feature.

### Use case

I'm creating a feature which allows users to suggest changes to translations. In order to know items which have been translated, the text must be wrapped in some kind of reference-able DOM element with a corresponding translationId.

### DIFF
--- a/docs/content/guide/de/15_custom-interpolators.ngdoc
+++ b/docs/content/guide/de/15_custom-interpolators.ngdoc
@@ -20,7 +20,7 @@ die ein Objekt zurück gibt, die folgendes Interface implementiert:
 
 - `setLocale(langKey)` - Setzt die Sprache mit Sprachschlüssel
 - `getInterpolationIdentifier()` - Gibt einen Identifier für den Service zurück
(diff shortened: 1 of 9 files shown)


joelongstreet:
What do you all think?

knalli:
Thank you, I think that is a very nice addition, but had no time to look deeper into it. Sorry :/

joelongstreet:
Sounds good. I went ahead and backed out the changes for the tests. Is this what you were asking for?

knalli:
Yes. But it would be nice if at least one test ensures the additional param will be transported correctly. The idea behind the test was basically fine, only the api break not :)

joelongstreet:
I'm sorry. I don't understand how it's possible to write a test which verifies a translationId returns as expected if only the string of a translation (and not the translationId) returns. Can you provide an example please.

knalli:
I've added one.

knalli:
Thank you for contribution. curated and landed as 5de40de

joelongstreet:
That's awesome! Thanks!

Issue #1657, opened 2016-12-02
Angular 1.6 support


alippai:
Bower says Angular 1.6 is incompatible with angular-translate 2.13.0:
`angular#>=1.2.26 <1.6 which resolved to 1.5.9 and is required by angular-translate#2.13.0`

alippai:
I wanted to add a PR as well, but I can't find the given constraint. :S

knalli:
Hi,

exactly this constraint is in the shadow repo only.

However, before adding this we have to ensure we are actually compatible with 1.6. It is not adding a version only. That is not so easy. If you know what you are doing, you can override this version on your own always any time.

The test targets are managed by:
* existence of a directory here: https://github.com/angular-translate/angular-translate/tree/46ba474b9be218ed26ab71822f398cb5409440c8/test_scopes
* an entry for TravisCI here https://github.com/angular-translate/angular-translate/blob/46ba474b9be218ed26ab71822f398cb5409440c8/.travis.yml#L20-L25

knalli:
Fine, tests added but there is something failing. That's why we don't open the version range in advance.

However, you can always override the version with bower resolutions.

alippai:
It's stable now: https://github.com/angular/angular.js/blob/master/CHANGELOG.md
Maybe the release notes and breaking changes will help to identify the problem.

evilaliv3:
@knalli any new on this front?

knalli:
Sorry, no time yet. PR is still open, also for anyone helping outstanding issues.

spock123:
The issue is with the $http service breaking changes

knalli:
The PR #1659 is still open. The issue is actually a promise rejection check (introduced in AJS 1.6). If someone can help out, that would be fine. Make the things faster.

z3ntu:
Can I expect this to be fixed in the near future?

knalli:
Well, #1659 is still open for everyone.

I'd sit down at this yesterday, but don't have a solution. Maybe it's a test scenario issue only (with custom $httpBackend.whenAsync implementation).

knalli:
in canary landed

evilaliv3:
cool @knalli thanks!

is there any ETA for the release?

knalli:
-> already released

PR #1659, opened 2016-12-06
feat(tests): add current AngularJS 1.6 in test scopes

knalli:
Relates #1657

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,4 @@ env:
   - TEST_SCOPE=angular_1.3.x
   - TEST_SCOPE=angular_1.4.x
   - TEST_SCOPE=angular_1.5.x
+  - TEST_SCOPE=angular_1.6.x
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -83,7 +83,10 @@ module.exports = function (grunt) {
         ]
       },
 
-      test: ['test/**/*.js']
+      test: [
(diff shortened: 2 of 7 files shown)


knalli:
Ok, mainly the issue is `Possibly unhandled rejection`. Although we could perhaps disable the check, this issue is still there: Where the heck is the missing promise handle?

knalli:
If we disable the checks [here](https://github.com/angular-translate/angular-translate/blob/37146d459be8d151379099d36ee4aae787125619/test/unit/service/translate.spec.js#L1035-L1036), the error messages will be less verbose but still there.

Regarding the unhandled rejection: I'm not sure.
* Looks like [this test](https://github.com/angular-translate/angular-translate/blob/37146d459be8d151379099d36ee4aae787125619/test/unit/service/translate.spec.js#L1071) is enough to run (replace temporarily locally `it` => `fit`).
* Anywhere in the chain starting [here](https://github.com/angular-translate/angular-translate/blob/37146d459be8d151379099d36ee4aae787125619/src/service/translate.js#L1888) is lacking a rejection handler. If [this](https://github.com/angular-translate/angular-translate/blob/37146d459be8d151379099d36ee4aae787125619/src/service/translate.js#L1895) is rejection an error (let's say `throw 'XYZ'`), then `XYZ` will be the unhandled rejection message.

Issue #1642, opened 2016-11-07
Error installing with Yarn (node >=6.9 required)


ashleymoogle:
### Subject of the issue
Since 2.13, engine in the package.json as been set to node 6.9 (released in october...). Had to update the global version of node.js.
It makes yarn fail the install and throw this error : 

yarn

yarn install v0.16.1
info No lockfile found.
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error angular-translate@2.13.0: The engine "node" is incompatible with this module. Expected version ">=6.9".
error Found incompatible module
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
 

npm ERR! Darwin 16.1.0
npm ERR! argv "/usr/local/Cellar/node/6.8.0/bin/node" "/usr/local/bin/npm" "run" "yarn"
npm ERR! node v6.8.0
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! 1001backend@0.0.0 yarn: `yarn`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the 1001backend@0.0.0 yarn script 'yarn'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the 1001backend package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     yarn
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs 1001backend
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls 1001backend
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/romainb/Code/1001backend-dev/npm-debug.log

npm ERR! Darwin 16.1.0
npm ERR! argv "/usr/local/Cellar/node/6.8.0/bin/node" "/usr/local/bin/npm" "run" "super-install"
npm ERR! node v6.8.0
npm ERR! npm  v3.10.8
npm ERR! code ELIFECYCLE
npm ERR! 1001backend@0.0.0 super-install: `npm run yarn-install && npm run yarn`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the 1001backend@0.0.0 super-install script 'npm run yarn-install && npm run yarn'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
[...]

PR #1660, opened 2016-12-06
chore(package): remove specific NodeJS version requirement

knalli:
This drops the explicit package’s `engine` field for the NodeJS version.
Instead, an additional `devEngines` has been added and will be forced verified with a custom dev-checker.

Solves #1642


knalli:
f74dd9a

PR #1664, opened 2016-12-13
fix(translate): handle null translation

### LINKED ISSUE #665: Null translations
Can we support null values in translations? For know:

``` javascript
  var translationMock = {
    'EXISTING_TRANSLATION_ID': 'foo',
    'BLANK_VALUE': '',
    'TRANSLATION_ID': 'Lorem Ipsum {{value}}',
    'TRANSLATION_ID_2': 'Lorem Ipsum {{value}} + {{value}}',
    'TRANSLATION_ID_3': 'Lorem Ipsum {{value + value}}',
    'NULL_TRANSLATION': null,
    'DOCUMENT': {
      'HEADER': {
        'TITLE': 'Header'
      },
      'SUBHEADER': {
        'TITLE': '2. Header'
      }
    }
  };

  describe('$translate', function () {
    it('should translate null values', function () {
      var deferred = $q.defer(),
          promise = deferred.promise,
          value;

      promise.then(function (translation) {
        value = translation;
      });

      $q.all([
        $translate("NULL_TRANSLATION")
      ]).then(function (translations) {
        deferred.resolve(translations);
      });

      $rootScope.$digest();
      expect(value[0]).toEqual('');
    });
  });
```

``` bash
Chrome 36.0.1985 (Linux) pascalprecht.translate $translate should translate null values FAILED
        TypeError: Cannot read property 'substr' of null
            at determineTranslation (/home/yanoo/Work/angular-translate/src/service/translate.js:1134:26)
            at $translate (/home/yanoo/Work/angular-translate/src/service/translate.js:781:11)
            at null.<anonymous> (/home/yanoo/Work/angular-translate/test/unit/service/translate.spec.js:277:9)
Chrome 36.0.1985 (Linux): Executed 251 of 251 (1 FAILED) (1.195 secs / 0.877 secs)
```

GuyMograbi:
null translation treated as if does not exist

fixes #665

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1612,7 +1612,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
       }
 
       // if the translation id exists, we can just interpolate it
-      if (table && Object.prototype.hasOwnProperty.call(table, translationId)) {
+      if (table && Object.prototype.hasOwnProperty.call(table, translationId) && table[translationId] !== null) {
         var translation = table[translationId];
 
         // If using link, rerun $translate with linked translationId and return it
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1353,7 +1353,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
       var deferred = $q.defer();
 
       var onResolve = function (translationTable) {
-        if (Object.prototype.hasOwnProperty.call(translationTable, translationId)) {
+        if (Object.prototype.hasOwnProperty.call(translationTable, translationId) && translationTable[translationId] !== null) {
           Interpolator.setLocale(langKey);
           var translation = translationTable[translationId];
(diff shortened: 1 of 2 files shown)


knalli:
Hello, 

thank you for this great contribution.

Technically, this PR is not ready as adding a feature for `$translate.instant(…)`, but not `$translate(…)`.

Additionally, I'm not sure if it would not be easier for you to create an own customer loader or post processor which removes null properties.

GuyMograbi:
Thanks so much for pointing out I should add the same for $translate. done. 

Regarding your loader/processor comment - I just don't want translate to break. currently it breaks on null values. 
So I want to define some default behavior, and if someone wants to change that behavior they can use loaders and processors. 

Currently it feels too much like a bug.

But I admit this is a matter of taste.

knalli:
Thank you, picked as 1e57b4f

GuyMograbi:
Hi, what does that mean? Will is be accepted or is it rejected?

PR #1674, opened 2017-01-13
Bugfix duplicate+parallel http requests per part

niaher:
If the same part is requested multiple times in quick succession, then multiple parallel HTTP requests will be fired. The fix here, makes sure that there's only 1 HTTP request per part.

### Description

💔Thank you!

### DIFF
--- a/src/service/loader-partial.js
+++ b/src/service/loader-partial.js
@@ -55,7 +55,7 @@ function $translatePartialLoader() {
     if (!this.tables[lang]) {
       var self = this;
 
-      return $http(angular.extend({
+      this.tables[lang] = $http(angular.extend({
         method : 'GET',
         url: this.parseUrl(urlTemplate, lang)
       }, $httpOptions))
@@ -75,7 +75,7 @@ function $translatePartialLoader() {
             return $q.reject(self.name);
           }
         });
-
+      return this.tables[lang];
     } else {
       return $q.when(this.tables[lang]);
     }


mtwar:
The change you're attempting here will cause an unexpected behavioral change because no subsequent call will ever create a new request, regardless of the success or failure of the original call, whereas the original code would try again. If the code using the partial loader has its own logic to retry the part registration/fetch (perhaps because it has some server status monitoring), then your implementation will cause a regression. 

I propose a more extensive change to chain promises for subsequent calls. Such an implementation will allow subsequent calls to execute a retry on failure, but no parallel HTTP requests will be made and success will pass through without extra requests. This is in-line with the intent of the original logic, which prevents duplicate requests on success. I made this change for our app but have not had an opportunity to resolve why I can't push to this repository yet; here is a diff with my changes: https://www.diffchecker.com/3CroelqL

knalli:
Thank you both. It will take some time looking into it.

knalli:
Sorry for the delay, but it has taken a while.

The only change is that the result of the complete request incl. post processing (say line 63 `self.tables[lang]`) will be stored in `this.tables[lang]`. (Note: `this` and `self` are the same, so the change should be using the alias `self` avoiding any confusion).

So, we have three scenarios:

1. The request is be successful (calling `then()`): The change will override the already existing data because `this.tables[lang]` would have been refreshed already.
2.
   1. The request is not successful and an errorHandler is given: On a successful alternative response, the tables will be refreshed with new data. Otherwise see 2.2
   2. The request is not successful and no errorHandler is given. No update of the tables.

But as a result of this change, always there will be a promise stored in `tables`where it should not be. And all others do not expect a promise in a table.

knalli:
Process of loading fixed, replaced by #1692

PR #1679, opened 2017-01-24
Just fixing a typo in a comment.

istro:
While looking to bust cache in our project, ended up reading the source of `$translate.refresh()` method. Noticed a typo, so here is a minute and insignificant fix.


knalli:
Thank you :)

PR #1682, opened 2017-01-27
Added keywords

naperkovskiy:
I'v noticed a keywords missing from package.json and added some.


knalli:
Thank you, landed in 64a2353

We have already keywords in the shadow repos, but missed them in angular-translate actually. 🚶

PR #1687, opened 2017-02-07
Prevent duplicate requests in partial loader

mtwar:
I updated the getTable function to chain subsequent calls off of the original HTTP request, passing through the value on success and reattempting the HTTP request on failure. This is pretty well described in my commit message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,28 @@
+<a name="2.9.0"></a>
+# [2.9.0](https://github.com/angular-translate/angular-translate/compare/2.8.1...v2.9.0) (2016-01-24)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


mtwar:
My commit list got a little messy from how I forked/updated this, but it's just my work in the diff :)

mtwar:
I see I'm getting test failures because of the use of "this", but the original function also used "this" and I'm handling it correctly with closure scoping. Should I be concerned about this?

knalli:
Do you think the check in jshint is incorrect?

mtwar:
I refactored it, will resolve the test failures ASAP.

mtwar:
@knalli this last test appears to be failing due to the intended behavioral change, which is that duplicate HTTP requests should not be made. The test is written assuming that five HTTP requests will be made for the three parts in its loader, whereas only three HTTP requests are now made. Should I just rewrite the test to account for this?

knalli:
Well.

1. Please rebase your commits against the latest canary, and squash your commits into a single or so commit. 
2. Regarding the tests: I'm not in this topic right now, I've to have a deeper look. If you want you can add copies of the tests in question and apply the intended demo case.

PR #1692, opened 2017-02-08
fix($translatePartialLoader): prevent duplicate simultaneous HTTP requests

mtwar:
The logic for getTable tries to prevent duplicate requests by checking if the data has been assigned or not, but this allows duplicate requests to execute if the first request has not even finished yet. When using the partial loader to load strings files per-control, this means that many dozens of requests may be made for a part associated with a low-level component that is reused across the page. My solution is to chain the promises for each request, allowing duplicate calls to execute against the server if all previous calls for that part in that language fail (allowing external retry logic to still work as it did before), but on success of any request the result will just be passed through the promise chain (like a bucket brigade) and every subsequent call will resolve with the requested data without trying to fetch it again. To the caller it is completely irrelevant that there is not actually another call made - their logic will execute off of the dependent promise just as it would off a new HTTP request. This is a more intuitive way of seeing the program behavior from a debugger than just returning the original promise because you will only see the handlers registered against any particular call to getTable in each promise's handler queues, rather than the entire set of handlers associated with the single server call as if the original promise were simply returned, so from the outside it is practically invisible that only a single server call is being made and the logic attached to a particular call is encapsulated in its own queue. The only behavioral change (besides the intended prevention of duplicate HTTP requests) is that duplicate calls to getTable will actually have their promise.then handlers executed serially in the order the calls were made, whereas multiple HTTP requests would result in undefined execution order dependent on the server response time for each thread. A unit test was also rewritten to account for the new behavior and checks to see that only one request is made per part and language and that the correct data is returned on language change.

### DIFF
--- a/src/service/loader-partial.js
+++ b/src/service/loader-partial.js
@@ -27,6 +27,7 @@ function $translatePartialLoader() {
     this.isActive = true;
     this.tables = {};
     this.priority = priority || 0;
+    this.langPromises = {};
   }
 
   /**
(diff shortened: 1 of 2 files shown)


knalli:
Curated, and landed as 8b2cea8

@mtwar  Very, very great contribution and a big thank you! 🎉

knalli:
Ping: Looks like something has broken, see #1703

PR #1693, opened 2017-02-08
fix($translate): reassign language promises in refresh, update translation tables at the appropriate time, and simplify the routine

mtwar:
When using $translatePartialLoader, the language promise is invalid upon adding a new part and calling $translate.refresh; calling $translate immediately after adding a new part results in the language key being returned from the already-resolved promise that was created on the original invokation of loadAsync, rather than waiting for the new part to be fetched from the server. Calling $translate.use with $forceAsyncReloadEnabled=true configured is a workaround with the current code, but it is logically incorrect to force a language change when a refresh is the appropriate action. This change remedies the problem by assigning the promise from loadAsync to the appropriate langPromise upon refresh.

With this fix in-place, it was revealed that that the asynchronous processing queue (aka $q, promise) generated by aggregating the results of the individual language promises in the code branch where all languages are refreshed executes the logic to update the translation tables after all of the individual table promises are already resolved and their queues have been processed. This means that, when using the partial-loader, any calls to $translate for strings contained in a newly-added part that are made before the entire refresh has finished will erroneously have a cache miss because the determineTranslation call is waiting on the individual table promises, which resolve all of their handlers (promise.then) before the outer handler ($q.all().then) begins processing its handler to actually update the translation tables. This is resolved by moving the translation table update handler to the individual event queues for each table, thereby guaranteeing the table has been updated by the time determineTranslation is executed because the handlers are executed in the order they are registered and the table-updating handler will be the first thing registered. After really digging in and understanding refresh I also noticed that there was some duplicate and overly-complex logic, so I refactored it to be a little simpler and shorter.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -2032,68 +2032,66 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
         throw new Error('Couldn\'t refresh translation table, no loader registered!');
       }
 
-      var deferred = $q.defer();
+      $rootScope.$emit('$translateRefreshStart', {language: langKey});
 
-      function resolve() {
-        deferred.resolve();
-        $rootScope.$emit('$translateRefreshEnd', {language : langKey});
+      var deferred = $q.defer(), updatedLanguages = {};
+
+      //private helper
+      function loadNewData(languageKey) {
+        var promise = loadAsync(languageKey);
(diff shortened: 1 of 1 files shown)


knalli:
@mtwar  Thank you for this nice additions!

Great improvements. Landed as 351eb8f

Yep, another gif.

PR #1731, opened 2017-05-08
Remove extra quotes in example

autowp:


knalli:
Thank you, landed as 69c666a

PR #1748, opened 2017-06-09
fix: Timezone and DST agnostic Unit test

TheTalisman:
Date generated for use in the unit test ‘should ignore a date param’ changed to be date only, in local time, to avoid having the test fail based on the timezone/DST of the machine the test is run on.

### Description
This fixes a bug that I encountered where the mentioned unit test was failing due to my machine's timezone offset. The evaluated date was off by one day from the provided (hardcoded) test value. As a note, my current timezone as of the fails is PDT. I tested this fix by running it many times and changing my local machine's timezone all around the world.

💔Thank you!


knalli:
Thank you. In which situation you got this? I wonder because two different people (you know... ;)) have been faced with this in a same time frame.

PR #1749, opened 2017-06-13
fix(unit test)

evancarey:
Call to new Date with ISO8601 string w/o time component yields UTC tz
date object which can be different from the expected date, '2016-08-21',
defined in the unit test. Calling new Date with arguments for year,
month, day will by pass the implicit call to Date.parse and avoid the
potential tz difference.

For example, when running the unit tests with new Date('2016-08-21') I would see the following output:

HeadlessChrome 0.0.0 (Mac OS X 10.12.5) pascalprecht.translate $translateDefaultInterpolation#interpolate should ignore a date param FAILED
	Expected 'Day is: 20.08.2016' to be 'Day is: 21.08.2016'.
	    at Object.<anonymous> (test/unit/service/default-interpolation.spec.js:150:72)
	    at Object.invoke (bower_components/angular/angular.js:3967:17)
	    at Object.workFn (bower_components/angular-mocks/angular-mocks.js:2177:20)

It is possible to use an ISO8601 date w/ time, '2016-08-21T00:00:00', but that also failed for me depending on what version of Chrome I had (v58 failed, v59 win). Date.parse called with an ISO8601 date string will create a date object with different tz depending on the components included in the date string. Avoiding the implicit call to Date.parse seemed to be the most predictable solution.

### DIFF
--- a/test/unit/service/default-interpolation.spec.js
+++ b/test/unit/service/default-interpolation.spec.js
@@ -141,7 +141,7 @@ describe('pascalprecht.translate', function () {
     it('should ignore a date param', inject(function ($translateSanitization) {
       var text = 'Day is: {{day | date:"dd.MM.yyyy"}}';
       var params = {
-        day : new Date('2016-08-21')
+        day : new Date(2016, 7, 21)
       };
       var sanitizedText = 'Day is: 21.08.2016';
 


TheTalisman:
LOL, exactly the same as my solution... fix: Timezone and DST agnostic Unit test #1748

evancarey:
@TheTalisman Yeah, I should a checked before going through the trouble of submitting the pull request. Didn't see yours until after.

TheTalisman:
@evancarey Yeah, easy to have happen.

I was just curious to see if you had a different solution =)

knalli:
@evancarey Thank you. Sorry about that, but #1748 was first and also made by a first-time contributor like you. 🐎 

I have asked it already: How does have this triggered initially? Because we have this tests for years and it has made never problems until now..

evancarey:
@knalli dunno.  What's the tz of your build/test server? Can't say why it's a new problem because I'm a new user of your lib.

knalli:
Well, it's running for years at Travis, I would guess they are using UTC. I'm not in UTC, so at least 2 different zones have worked. Strange..

evancarey:
Are you East of UTC? I'm West by 5hrs (CDT). Wonder if my being -5hrs from UTC is what puts my localized date object's day value into the previous day? 2016-08-21T00:00:000Z in my tz is Aug 20 2016 19:00:00 GMT-0500 (CDT). I'm guessing that if I were East of UTC by a couple of TZs the localized date would still come out as Aug 21.

TheTalisman:
Like @evancarey, I ran into this with my first pull and build attempt. So, I have no insight into any over time changes to the behavior. But, I'm pretty sure that there was at least the possibility repro'ing this for quite some time.

As I believe that I noted in my PR, the timezone that I was running under when I initially encountered it was PDT. I ran a bunch of validation tests, changing my machine's location all around the world and the running the tests to verify the fix.

Out of curiosity, @evancarey, what TZ were you running under on the machine your repro'd this?

Anyway, the problem would vary because the test date was generated as a UTC date while an unqualified `new Date()` created in Chrome is generated as a local date. Since only the _date_ portion of the timestamp was being compared, in order for a repro to occur the build machine had to be in a TZ where the conversion from UTC to local resulted in the hours being shifted _enough_ to push the date portion of the timestamp into a different day. If you test the previous version, and set the local machine to different timezones around the world, you'll see that some locations will repro while others won't, because their local time offset against UTC is not enough to change the day (this will depend on the current time, too). As an aside, to make matters a little more complicated, Chrome's default Date usage used to be UTC and they switched it to local time, I'm not sure though what rev that happened in =)

knalli:
> Are you East of UTC? I'm West by 5hrs (CDT). Wonder if my being -5hrs from UTC is what puts my localized date object's day value into the previous day? 2016-08-21T00:00:000Z in my tz is Aug 20 2016 19:00:00 GMT-0500 (CDT). I'm guessing that if I were East of UTC by a couple of TZs the localized date would still come out as Aug 21.

Ah, gotcha. :)

PR #1751, opened 2017-06-14
fix(service): fixed IE8 "Expected identifier" error

vgoma:
"catch" and "finally" is reserved words in IE8. It produces 'Expected identifier' errors. Fixed that with quotes.


knalli:
@vgoma  thank you for contributing

PR #1760, opened 2017-07-12
feat(loader-partial): addPart specific urlTemplate override

evancarey:
To accommodate use case in which location of translation table for a
partial is not known at config time and is unique to partial being
requested. For example, app is served as a SPA from s3 bucket, after
initialization and based on route parameters addParts from other s3
buckets and/or paths based on those route parameters.

### DIFF
--- a/src/service/loader-partial.js
+++ b/src/service/loader-partial.js
@@ -22,12 +22,13 @@ function $translatePartialLoader() {
    * @description
    * Represents Part object to add and set parts at runtime.
    */
-  function Part(name, priority) {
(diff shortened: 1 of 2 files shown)


knalli:
@evancarey  Thank you for contributing, a nice addition.

Picked as 633fbc9

yrtimiD:
[The main documentation site](https://angular-translate.github.io/docs/#/api/pascalprecht.translate.$translatePartialLoader) doesn't have docs for this new parameter. Can anybody update it?

Issue #1781, opened 2017-09-27
PartialLoader: $translate.use() and $translate.refresh() in race-condition, wrong language loaded


SimonSimCity:
### Subject of the issue
If I call `$translate.use()` and add another partial template directly after that, this one won't get loaded in the new language. 

### Your environment
* version of angular-translate: 2.15.2
* version of angular: 1.5.11
* which browser and its version: Every ...

### Steps to reproduce

https://www.webpackbin.com/bins/-Kv2DNtdpnsNYAU6gMTA

1. Set up a project using the [`partialLoader`](https://angular-translate.github.io/docs/#/guide/12_asynchronous-loading) having at least two languages (f.e. English and German).
1. Set English as fallback language
1. Call `$translate.refresh()` every time your parts change:
```js
  $rootScope.$on('$translatePartialLoaderStructureChanged', function () {
    $translate.refresh();
  });
```
4. Let your app start off by choosing German as language (by `$translate.use('de')`) and in the next line add a new part.

### Expected behaviour
The new part should be available in German

### Actual behaviour
The new part is only loaded in English, the old (in this case the fallback) language.

### Explanation
When calling `$translate.use('de')`, the language effectively hasn't changed, but it has a callback, that is called when the change of the language was successful. This creates a problem in this case here, because when now calling `$translatePartialLoader.addPart('test')` and thereby `$translate.refresh()`, before the callback of `$translate.use('de')` comes back, the value of the internal variable `$uses` is still on the previous value, and therefore the call of `$translate.refresh()` is called before the value is changed. This method in turn will reload all the available translations after emptying them. This will also overwrite the content of the `$translate.use('de')` callback by the time it comes back.

### A workaround
Call `$translate.refresh()` again after `$translation.use('de')` has finished. Use the promise like shown here:
```js
$translation.use('de').then(() => $translate.refresh());
```

The question of me now:

Does all this `$translate.refresh()` pull down on my performance? I haven't run tests against it, but I could well imagine that this (having quite some parts loaded) can take on performance ...

SimonSimCity:
Could well be a duplicate of #1704, #1703 ... The issue #1703 can at least be solved by the work-around I provided here.

knalli:
@SimonSimCity  We could consider to call such a refresh implicitly (for the user). Did you have see some (performance) issues with this? 

Maybe you want make a PR for this?

knalli:
[...]

PR #1805, opened 2017-11-02
Refresh translations after changing it. This is a work-around to #1781

SimonSimCity:
Implemented a work-around as described in #1781. Not very proud of it, but it works.

A propper fix would be to change the language before loading the async parts, whereby also the rest for #1781 should work.

This work-around get's many tests to crash, but I don't have the insight into the project in order to fix them. I'd appreciate it if someone could it pick up from here.

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -1945,6 +1945,8 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
         useLanguage(key);
       }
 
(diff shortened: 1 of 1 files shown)


knalli:
@SimonSimCity  Thank you for contribution.

Well, this is not working (obviously). The "workaround" is active always, not only for partial loaded content.

I have prepared something, see #1781

SimonSimCity:
I guess I underestimated it pretty much 😄  I didn't put much time into it and was therefore only focused on the bug I have instead of looking for a solution that helps the plugin as unit.

PR #1821, opened 2018-01-24
feat(*): support google closure compiler

maleetz:
Because of duplicate definitions of getTranslateNamespace in the same scope, google
closure compiler returns the following error when minifying the angular-translate.js library:
- Duplicate let / const / class / function declaration in the same scope is not allowed.
Change one of the getTranslateNamespace to be _getTranslateNamespace to avoid error.

🚨Please review the [guidelines for contributing](../CONTRIBUTING.md) to this repository.

### Description

💔Thank you!

### DIFF
--- a/src/directive/translate-namespace.js
+++ b/src/directive/translate-namespace.js
@@ -60,7 +60,7 @@ function translateNamespaceDirective() {
     compile: function () {
       return {
         pre: function (scope, iElement, iAttrs) {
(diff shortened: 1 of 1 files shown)


knalli:
Well: I would skip such checks and would not put ES6+ over this code.. we have simply no support for this. No idea what will happen in the future.

However, this change is okay. Thank you for contribution!

PR #1891, opened 2019-12-17
feat: throw TypeError for translationId

gabbsmo:
🚨Please review the [guidelines for contributing](../CONTRIBUTING.md) to this repository.

### Description
AngularJS 1.6 introduced a new behaviour that caused an exception on unhandled rejections of promises. If you were to pass undefined to to $translate the error message would simply be "Possibly unhandled rejection: undefined". With this change a TypeError is thrown instead so that you can easly find where the invalid translationId was passed from in the stack trace

💔Thank you!


knalli:
@gabbsmo Thank you for contribution! Picked as b363e3b

PR #1910, opened 2021-01-15
feat(service): add allowNamespace parameter

frenautvh:
Hi,

I'm using angular-translate on an application with lots of large translation json files.

While profiling the application using chrome performance tool i noticed that the multiple calls to the "flatObject" method is taking a whole second of processing time, which is very costly.
Since i'm not using namespaces at all, flatObject iterates over all the key/values for nothing.

I suggest to add the ability to simply disable the usage of namespaces if you don't need them (by adding an allowNamespaces parameter). In my case, disabling it gives me a noticeable performance boost (1s gained since the flatObject processing is just ignored).

### DIFF
--- a/src/service/translate.js
+++ b/src/service/translate.js
@@ -36,6 +36,7 @@ function $translate($STORAGE_KEY, $windowProvider, $translateSanitizationProvide
     $notFoundIndicatorRight,
     $postCompilingEnabled = false,
     $forceAsyncReloadEnabled = false,
+    $allowNamespaces = true,
     $nestedObjectDelimeter = '.',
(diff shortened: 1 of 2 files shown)


knalli:
Hi @frenautvh ,

thank you for contributing. Although you waited patiently, at the end we have missed it. Out of my radar, sorry about this. As this is still valuable addition, I will open it again. 

Thank you.

qpic-cgi (approved):

knalli:
Omg.. thank you again for the reminder :)

2.19 is available with your additional now. 🐤

